PR #2, opened 2023-06-06
Chunking RFC

mrmr1993:
Fix https://github.com/MinaProtocol/mina/issues/12880


0002-chunking.md
```diff
@@ -0,0 +1,271 @@
+## Summary
```
barriebyron, line 1:
```suggestion
## Chunking summary
```

barriebyron: I suggest we change "Summary" to the gist of the RFC so we can identify the subject matter at a glance, otherwise, all of our RFCs will be titled "Summary" (not very useful)

0002-chunking.md
```diff
@@ -0,0 +1,271 @@
...
+Finally, we can compute the value `f(z)` by proxy:
+```text
+f(z) = f_0(z) + z^n f_1(z) + z^{2n} f_2(z) + z^{3n} f_3(z)
+```
+
+Thus, this chunking approach provides all of the features that we need to support the kimchi proof system over a larger domain size than the current Mina limit.
+
+### Retaining zero-knowledge
+
+Once we have `c > 1` chunks, every column will come with `2 * c` evaluations, 1 for each chunk at each of `z` and `z * w`. In order to retain zero-knowledge, we need to ensure that we randomize more columns than there are evaluations. Thus, we must randomize at least `2 * c + 1` values at the end of each column. The helpers for adding zero-knowledge to columns should be updated to take the number of zk rows that are required as an argument.
```
barriebyron, line 83:
```suggestion
After we have `c > 1` chunks, every column will come with `2 * c` evaluations, 1 for each chunk at each of `z` and `z * w`. In order to retain zero-knowledge, we need to ensure that we randomize more columns than there are evaluations. Thus, we must randomize at least `2 * c + 1` values at the end of each column. The helpers for adding zero-knowledge to columns must be updated to take the number of zk rows that are required as an argument.
```

barriebyron: if we mean "must" let's be sure to use must, otherwise I am reading "should" as a suggestion

0002-chunking.md
```diff
@@ -0,0 +1,271 @@
...
+`zk_rows` and `domain_size` satisfying the fixpoint
+
+```text
+zk_rows = (16 * (domain_size / max_poly_size) + 5) / 7
+domain_size = circuit_size + zk_rows
+```
+
+### Technical approach in kimchi
+
+Support for chunked commitments already exists [in kimchi's supporting polynomial commitment library](https://github.com/o1-labs/proof-systems/blob/ca0b9e53d15e3ab5db6fb8c20617979cbf74215e/poly-commitment/src/srs.rs#L89). This RFC proposes to extend that support through to the kimchi prover, where an attempt to use the feature currently [results in an error](https://github.com/o1-labs/proof-systems/blob/ca0b9e53d15e3ab5db6fb8c20617979cbf74215e/kimchi/src/prover.rs#L170). This RFC also proposes to add support for chunked commitments and evaluations to pickles, where there is currently no support.
```
barriebyron, line 192: I feel like we "buried the lede" here, this detail would be very helpful in RFC summary section

0002-chunking.md
```diff
@@ -0,0 +1,271 @@
...
+
+The verifier logic for [the public input commitment](https://github.com/o1-labs/proof-systems/blob/ca0b9e53d15e3ab5db6fb8c20617979cbf74215e/kimchi/src/verifier.rs#L592) will need to use the chunked lagrange basis, to be consistent with how we will generate them on the prover side.
+
+To avoid the `O(c n log(n))` cost of computing the chunked evaluation for the public input polynomial in the verifier, the prover should pass this evaluation as part of the proof. In order to avoid breaking compatibility with zkApps, this 'public input' field should be optional, as proofs submitted in zkApp commands will not have this field available. This value is already considered part of the transcript for the purposes of pickles, so there is no soundness risk of using a value provided by a malicious prover.
+
+Once all of the above are complete, the error checks in kimchi's prover and verifier can be removed, and chunking can be considered enabled in kimchi. This should be accompanied by a test that exercises this behavior.
```
barriebyron, line 202:
```suggestion
After all of the above changes are complete, the error checks in kimchi's prover and verifier can be removed, and chunking can be considered enabled in kimchi. This should be accompanied by a test that exercises this behavior.
```

0002-chunking.md
```diff
@@ -0,0 +1,271 @@
...
+
+### Technical approach in pickles
+
+The pickles code should be updated to reflect the changes to the bindings, which will propagate `array`s into several types. The logic for handling these will then need to be updated to handle these cases, in particular the [`ft_comm` calculation](https://github.com/MinaProtocol/mina/blob/c71d0f8852774b718a8f0dd5af37520985e57813/src/lib/pickles/common.ml#L231) will need to be updated to match [the kimchi logic](https://github.com/o1-labs/proof-systems/blob/ca0b9e53d15e3ab5db6fb8c20617979cbf74215e/kimchi/src/verifier.rs#L705), absorbing the evaluations in the [step verifier](https://github.com/MinaProtocol/mina/blob/c71d0f8852774b718a8f0dd5af37520985e57813/src/lib/pickles/step_verifier.ml#L928) and the [wrap verifier](https://github.com/MinaProtocol/mina/blob/c71d0f8852774b718a8f0dd5af37520985e57813/src/lib/pickles/wrap_verifier.ml#L871), and the `check_bulletproof` logic for [step](https://github.com/MinaProtocol/mina/blob/c71d0f8852774b718a8f0dd5af37520985e57813/src/lib/pickles/step_verifier.ml#L222) and [wrap](https://github.com/MinaProtocol/mina/blob/c71d0f8852774b718a8f0dd5af37520985e57813/src/lib/pickles/wrap_verifier.ml#L388).
```
barriebyron, line 212:
```suggestion
The pickles code must be updated to reflect the changes to the bindings, which will propagate `array`s into several types. The logic for handling these will then need to be updated to handle these cases, in particular the [`ft_comm` calculation](https://github.com/MinaProtocol/mina/blob/c71d0f8852774b718a8f0dd5af37520985e57813/src/lib/pickles/common.ml#L231) will need to be updated to match [the kimchi logic](https://github.com/o1-labs/proof-systems/blob/ca0b9e53d15e3ab5db6fb8c20617979cbf74215e/kimchi/src/verifier.rs#L705), absorbing the evaluations in the [step verifier](https://github.com/MinaProtocol/mina/blob/c71d0f8852774b718a8f0dd5af37520985e57813/src/lib/pickles/step_verifier.ml#L928) and the [wrap verifier](https://github.com/MinaProtocol/mina/blob/c71d0f8852774b718a8f0dd5af37520985e57813/src/lib/pickles/wrap_verifier.ml#L871), and the `check_bulletproof` logic for [step](https://github.com/MinaProtocol/mina/blob/c71d0f8852774b718a8f0dd5af37520985e57813/src/lib/pickles/step_verifier.ml#L222) and [wrap](https://github.com/MinaProtocol/mina/blob/c71d0f8852774b718a8f0dd5af37520985e57813/src/lib/pickles/wrap_verifier.ml#L388).
```

0002-chunking.md
```diff
@@ -0,0 +1,271 @@
...
+
+Due to the requirement for proof layout homogeneity, the `Pickles.compile` interface should take parameters for the `step_num_chunks` and `wrap_num_chunks` at the toplevel. These arguments may be optional labelled arguments, since they will not be regularly used in the protocol circuits. These arguments will have to be passed explicitly by the user / caller when needed, since it will be prohibitively expensive to try all of the various options programatically on every compile.
```
barriebyron, line 216:
```suggestion
Due to the requirement for proof layout homogeneity, the `Pickles.compile` interface must take parameters for the `step_num_chunks` and `wrap_num_chunks` at the top level. These arguments may be optional labelled arguments, since they will not be regularly used in the protocol circuits. These arguments will have to be passed explicitly by the user / caller when needed, since it will be prohibitively expensive to try all of the various options programmatically on every compile.
```

0002-chunking.md
```diff
@@ -0,0 +1,271 @@
...
+
+The `Pickles.compile` function should be modified to check for the number of chunks in each inductive rule's previous rule tags. When these are not consistent between inductive rules, the code must throw an error in order to achieve layout homogeneity.
```
barriebyron, line 218:
```suggestion
The `Pickles.compile` function must be modified to check for the number of chunks in each inductive rule's previous rule tags. When these numbers are not consistent between inductive rules, the code must throw an error in order to achieve layout homogeneity.
```

barriebyron: @mrmr1993  I love this first PRD, congratulations! What do we mean by "these" (these tags? or these numbers?)

barriebyron (commented):
Congratulations and kudos @mrmr1993  for submitted our first O(1) Labs official PRD.

Your PRD is a model of completeness, context, benefit, and problems solved. Well done!

0002-chunking.md
mrmr1993: @bkase wdyt for the template? Should we propose a structure that customises the summary for each?

0002-chunking.md
mrmr1993: `These numbers` is correct. The pickles recursion layer needs everything to have the 'same shape', but the 'number of chunks' result in different layouts for the proofs that it consumes. Is there a way to word this that would make that detail clearer?

0002-chunking.md
```diff
@@ -0,0 +1,271 @@
...
+
+### Prototype PRs
+
+* [Fix chunking for proofs](https://github.com/o1-labs/proof-systems/pull/1032)
+* [Update bindings for chunking fix](https://github.com/MinaProtocol/mina/pull/13285)
+* [Enable chunked proving](https://github.com/o1-labs/proof-systems/pull/1033)
+* [Add zero-knowledge for chunked proofs](https://github.com/o1-labs/proof-systems/pull/1045)
+* [Update bindings to enable chunking](https://github.com/MinaProtocol/mina/pull/13286)
+
+## Test Plan and Functional Requirements
```
bkase, line 228: This section needs a little more detail I think -- unit testing what specifically at each layer? Which parts of the system? And then we should sketch out the SnarkyJS test also in a bit more detail, but maybe that entails figuring out the SnarkyJS details first: I suppose something with ZkProgram though?

0002-chunking.md
```diff
@@ -0,0 +1,271 @@
...
+The motivation, assessment criteria, and exit criteria can be found in the [PRD](https://www.notion.so/minaprotocol/Chunked-Commitments-PRD-890c836529c545f082a3eeecd3d4f510).
+
+This proposal raises the maximum 'circuit size' limit for kimchi proofs from `2^16` to `2^32`. Out-of-scope is further work which would remove this limit, since this limit will not currently be reachable without other improvements to memory usage and performance.
+
+This proposal optimizes for compatibility with the kimchi and pickles [versions proposed for deployment in the Berkeley hard-fork](https://github.com/MinaProtocol/MIPs/blob/main/MIPS/mip-kimchi.md), as well as the [proposed zkApps feature](https://github.com/MinaProtocol/MIPs/blob/main/MIPS/mip-zkapps.md), at the expense of some potential loss of performance when the feature is in use.  Timing measurements can be found in the PR description for the [prototype implementation](https://github.com/o1-labs/proof-systems/pull/1033).
+
+## Detailed design
+[detailed-design]: #detailed-design
+
+### Theory overview
```
bkase, line 18: This is amazing content, but maybe it's better served as a chapter of the Mina book and linked to from this RFC, so it's more accessible? What do you think?

This is of course not clear from the current template, so we can adjust after we settle this.

0002-chunking.md
bkase: We could add a top-level `#` heading with the title of the feature maybe? But then keep `Summary` because it's a standard. What do you think @barriebyron ?

0002-chunking.md
```diff
@@ -0,0 +1,271 @@
...
+This simplifies to `k > (2 * c - 2) / 7`, giving `zk_rows > (16 * c - 2) / 7`.
+We can derive `c` from the `max_poly_size` supported by the URS, and thus we find
+`zk_rows` and `domain_size` satisfying the fixpoint
+
+```text
+zk_rows = (16 * (domain_size / max_poly_size) + 5) / 7
+domain_size = circuit_size + zk_rows
+```
+
+### Technical approach in kimchi
```
bkase, line 190: This part I think is definitely perfect to remain in the RFC

0002-chunking.md
```diff
@@ -0,0 +1,271 @@
...
+## Motivation
+[motivation]: #motivation
+
+The motivation, assessment criteria, and exit criteria can be found in the [PRD](https://www.notion.so/minaprotocol/Chunked-Commitments-PRD-890c836529c545f082a3eeecd3d4f510).
+
+This proposal raises the maximum 'circuit size' limit for kimchi proofs from `2^16` to `2^32`. Out-of-scope is further work which would remove this limit, since this limit will not currently be reachable without other improvements to memory usage and performance.
+
+This proposal optimizes for compatibility with the kimchi and pickles [versions proposed for deployment in the Berkeley hard-fork](https://github.com/MinaProtocol/MIPs/blob/main/MIPS/mip-kimchi.md), as well as the [proposed zkApps feature](https://github.com/MinaProtocol/MIPs/blob/main/MIPS/mip-zkapps.md), at the expense of some potential loss of performance when the feature is in use.  Timing measurements can be found in the PR description for the [prototype implementation](https://github.com/o1-labs/proof-systems/pull/1033).
+
+## Detailed design
```
bkase, line 15: Does this impact anything other than Pickles/Kimchi and SnarkyJS (and the bindings)? Probably not, but just double checking. Maybe worth calling out explicitly? I tried to do that in the RFC I just drafted.

0002-chunking.md
```diff
@@ -0,0 +1,271 @@
...
+## Prior art
+[prior-art]: #prior-art
+
+The existing broken chunking implementation.
+
+## Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+* What parts of the design do you expect to resolve through the RFC process before this gets merged?
+  - SnarkyJS design
```
bkase, line 267: What exactly do we need to expose to the user? The number of chunks? Can't we compute that at circuit compile time? This is important for the SnarkyJS API design

0002-chunking.md
barriebyron: better, yes. I updated the template, too

mitschabaude:
Amazing RFC, @mrmr1993! I was intrigued by this because I love Lagrange polynomials:

> Value columns in the Mina proof are represented in evaluation form, as a list of values. In order to support efficient evaluations over these values, the [LagrangeBasisEvaluations](https://github.com/o1-labs/proof-systems/blob/ca0b9e53d15e3ab5db6fb8c20617979cbf74215e/kimchi/src/lagrange_basis_evaluations.rs#L7) structure will need to be generalized to support evaluations of the 'chunked lagrange basis'. The approach to achieve this is equivalent to the one used in [the SRS's add_lagrange_basis function](https://github.com/o1-labs/proof-systems/blob/master/poly-commitment/src/srs.rs#L89), but using the explicit powers of z (z^0, z^1, ..., z^n) in place of the SRS curve points. This uses an [iFFT](https://en.wikipedia.org/wiki/Fast_Fourier_transform), so will result in a performance degradation from linear time (O(n)) to O(c n log(n)) when chunking is in use, where c is the number of chunks.

I hope I didn't misunderstand the problem statement, but I think I found a way to do polynomial evaluations in O(n).

So the problem, as I understood it, is to evaluate $f_0(z), ..., f_3(z)$ at an arbitrary point $z$ (sticking with your example of 4 chunks), given the evaluations of

$$f = f_0 + x^n f_1 + x^{2n} f_2 + x^{3n} f_3$$

over the domain $1,\omega,...,\omega^{4n-1}$. Here $\omega$ is a $4n$-th root of 1. 

The rough idea to solve it is to use this formula for the Lagrange polynomials:

$$L_i(x) = \frac{1}{4n} \sum_{j=0}^{4n-1} (\omega^{-i} x)^j$$

And now split up each Lagrange polynomial in a way that matches the splitting up of $f$ into $f_0,...,f_3$. In fact you can split it into

$$L_i(x) = L^0_i + (\omega^{-i}x)^{n}L^0_i + (\omega^{-i}x)^{2n}L^0_i + (\omega^{-i}x)^{3n}L^0_i$$

where

$$L^0_i(x) = \frac{1}{4n} \sum_{j=0}^{n-1} \omega^{-ij} x^j.$$

Plugging that into the evaluation formula for $f$ from its coefficients,

$$f(x) = \sum_{i=0}^{4n-1} f(\omega^i) L_i(x),$$

you arrive at corresponding formulas for the chunks $f_k$, $k = 0,1,2,3$,

$$f_k(x) = \sum_{i=0}^{4n-1} f(\omega^i) \omega^{-ikn} L^0_i(x).$$

All the $L^0_i(x)$ can be evaluated in ${O}(n)$, and so, all the $f_k$ can as well. In fact, if you split up $i$ by its residue modulo 4, like $i \rightarrow 4k + i$, I can show that

$$4 L^0_{4k + i}(x) = \frac{\omega^{4i}}{n} \prod_{j\ne i}^{n-1} (\omega^{-k}x - \omega^{4j}) = L^n_i (\omega^{-k}x)$$

where the right hand side is just the $i$-th Lagrange polynomial with respect to the domain $1,\omega^{4},...,\omega^{4(n-1)}$, generated by the $n$-th root of 1, $\omega^4$.

(As an aside, I think the [LagrangeBasisEvaluations](https://github.com/o1-labs/proof-systems/blob/ca0b9e53d15e3ab5db6fb8c20617979cbf74215e/kimchi/src/lagrange_basis_evaluations.rs#L7) can be improved to need fewer multiplications and no division.)

jspada (commented):
Really nice write-up.  Love the detail and the level it was written at.  I've asked some questions that occur to me, which may be useful to the reader if answered with some detail.

PR #4, opened 2023-06-06
edits for RFC 0001-template.md

barriebyron:
This PR:
- adds a level 1 heading for the RFC title to improve readability and enable quick document identification.
- minor grammatical changes for the template
- adds a description of prior art


0001-template.md
```diff
@@ -86,6 +87,6 @@ Discuss prior art, both the good and the bad, in relation to this proposal.
 ## Unresolved questions
```
barriebyron, line 87: what do we mean by "art" in:

> ## Prior art
> [prior-art]: #prior-art
> Discuss prior art, both the good and the bad, in relation to this proposal.

0001-template.md
```diff
@@ -1,12 +1,12 @@
...
 [summary]: #summary
 
-One paragraph explanation of the feature.
+One paragraph explanation of the feature. Describe the "why" and feature benefit.
 
 ## Motivation
 [motivation]: #motivation
 
-Reflect the will of the PRD (Product Requirements Document). If a linkable artifact, include it here. If not, summarize in your own words.
+Reflect the will of the PRD (product requirements document). If the PRD is a linkable artifact, include the link here. If not, summarize the product requirement in your own words. Describe the delta of this change.
```
barriebyron, line 9: Kudos to @mrmr1993  who provided a theory overview in our first RFC! https://github.com/o1-labs/rfcs/pull/2/files#diff-2c04d1657aa6e49b23d97667412e570417f73bc72b2f5b32dca24196832504c8R18 

should we can add suggestions here in the template to encourage this level of context?

0001-template.md
bkase: The term is taken from here: https://en.wikipedia.org/wiki/Prior_art (I think -- we took it from (a very old version of): https://github.com/rust-lang/rfcs/blob/master/0000-template.md )

0001-template.md
```diff
@@ -82,7 +82,13 @@ Why should we *not* do this?
 ## Prior art
 [prior-art]: #prior-art
 
-Discuss prior art, both the good and the bad, in relation to this proposal.
+Discuss prior art, both the good and the bad, in relation to this proposal. 
+
+Prior art is any evidence that your feature (invention, change, proposal) is already known. Examples include:
```
bkase, line 87: these examples are too coupled to Rust so they don't make sense in our context. I'd suggest:

```
Thinks about the lessons from other blockchain projects or similar updates and provide readers of your RFC with a fuller picture. If there is no prior art, that is fine. Your ideas are interesting whether they are new or adapted from another source.
```

and then just skip that inner section

barriebyron:
back at ya @bkase  

the suggestion for a top-level title makes so much sense... I know from experience that documents titled "Summary" are difficult to distinguish from each other.

bkase (approved):

PR #5, opened 2023-06-07
Adds Minimal SnarkyJS checks part1 RFC

bkase:


0003-minimal-snarkyjs-checks-part1.md
```diff
@@ -0,0 +1,91 @@
...
+Having said that, there are important reasons to keep some tests running on Mina's PRs: For example, many cryptography changes have the potential to cause proof generation on SnarkyJS, and by far the best place to address that is the PR introducing the cryptography change.
+
+As of this RFC, all the SnarkyJS are configured to "soft-fail" on Mina's CI; this was a quickfix to empower quick Mina development, but is creating opportunities for bitrot in these situations so this RFC aims to quickly reconcile this temporary situation.
+
+Success Metric: Protocol developers are undisturbed by SnarkyJS tests in day-to-day development, but when deep changes to cryptography are taken, 
+
+Requirements:
+
+* The bare-bones SnarkyJS tests (as described below) run as blocking jobs on Mina PRs
+* Any SnarkyJS tests that are more tightly coupled to zkApps do not block Mina CI
```
deepthiskumar, line 22: The unit tests in snarkyJS could still block Mina CI, right? Are they not dependent on mina types?

0003-minimal-snarkyjs-checks-part1.md
```diff
@@ -0,0 +1,91 @@
+## Summary
+[summary]: #summary
+
+A minimal layer of SnarkyJS checks that won't break with typical protocol development.
```
psteckler, line 4: Sentence fragment

How about:

"We describe a minimal layer of SnarkyJS check that won't hinder typical protocol development."

0003-minimal-snarkyjs-checks-part1.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Motivation
+[motivation]: #motivation
+
+Objective: Empower the protocol development to continue swiftly, while preventing maximally complex SnarkyJS bitrot. Out-of-scope for this RFC: re-enabling the rest of those tests in SnarkyJS CI, and changing the submodule layout within Mina.
+
+Background: Currently, SnarkyJS and, transitively, SnarkyJS-bindings is included in the main Mina repo as a submodule. Until recently, extensive CI jobs ensured that SnarkyJS remains correct as Mina evolves. Unfortunately, this setup has many issues: (1) Mina is owned by the Mina Foundation and SnarkyJS is owned by O(1) Labs; it's semantically awkward to rely on Mina's CI for testing SnarkyJS, (2) the dependancy graph feels a bit inverted; as SnarkyJS is a client of the Mina daemon's internals rather than Mina daemon's internals being a client of SnarkyJS, and most importantly (3) it's causing velocity issues with protocol development. Why? Many parts of SnarkyJS is tightly coupled to zkApp representations and behavior -- this means that even trivial renaming or reordering of fields requires SnarkyJS changes (albeit in a semi-automatic fashion).
```
psteckler, line 11: dependency

psteckler: "are tightly coupled"

0003-minimal-snarkyjs-checks-part1.md
```diff
@@ -0,0 +1,91 @@
...
+
+Having said that, there are important reasons to keep some tests running on Mina's PRs: For example, many cryptography changes have the potential to cause proof generation on SnarkyJS, and by far the best place to address that is the PR introducing the cryptography change.
```
psteckler, line 13: "to break proof generation" ?

psteckler (approved):
As an exercise for a new SnarkyJS test, I plan to unbox the new global slot types, which will change their serialization. That should break SnarkyJS, and the test should reveal that.

nholland94 (commented):
To help me understand the reason behind the approach, can we expand on why it makes sense to continue to keep the tests inside of the Mina repository? My question is 2-fold:

1. If we are just keeping the SnarkyJS test suite as a soft fail in the repo, isn't this strictly more difficult to deal with than inverting the submodules? Eg: The Mina protocol code and SnarkyJS code is out of date -- if I have Mina as a submodule to SnarkyJS, then I can diff the changes that caused the mismatches when I updated the submodule; if I have things as is, I have to go through the commit history, look at CI, see where things started failing, and then work through that diff.
2. What class of issues does the end-to-end recursive proof test prevent compared to the full SnarkyJS smart contract test suite?

Additionally, this document does not currently say anything about the Mina integration tests which rely on SnarkyJS. If those are still in and blocking, then schema changes will still break Mina's CI, since we are actually executing the smart contract code of SnarkyJS in those tests.

0003-minimal-snarkyjs-checks-part1.md
```diff
@@ -0,0 +1,91 @@
+## Summary
+[summary]: #summary
+
+A minimal layer of SnarkyJS checks that won't break with typical protocol development.
+
+## Motivation
+[motivation]: #motivation
+
+Objective: Empower the protocol development to continue swiftly, while preventing maximally complex SnarkyJS bitrot. Out-of-scope for this RFC: re-enabling the rest of those tests in SnarkyJS CI, and changing the submodule layout within Mina.
```
barriebyron, line 9:
```suggestion
Objective: Empower the protocol development to continue swiftly while preventing maximally complex SnarkyJS bitrot. Out-of-scope for this RFC: re-enabling the rest of those tests in SnarkyJS CI and changing the submodule layout within Mina.
```

0003-minimal-snarkyjs-checks-part1.md
barriebyron, line 11:
```suggestion
Background: Currently, SnarkyJS and, transitively, SnarkyJS-bindings are included in the main Mina repo as a submodule. Until recently, extensive CI jobs ensured that SnarkyJS remains correct as Mina evolves. Unfortunately, this setup has many issues: 

(1) Mina is owned by the Mina Foundation and SnarkyJS is owned by O(1) Labs; it's semantically awkward to rely on Mina's CI for testing SnarkyJS.
(2) The dependency graph feels a bit inverted, as SnarkyJS is a client of the Mina daemon's internals rather than Mina daemon's internals being a client of SnarkyJS.
(3) Most importantly, CI jobs are causing velocity issues with protocol development. Why? Many parts of SnarkyJS are tightly coupled to zkApp representations and behavior -- this means that even trivial renaming or reordering of fields requires SnarkyJS changes (albeit in a semi-automatic fashion).
```

barriebyron: let's improve readability with this suggested formatting

0003-minimal-snarkyjs-checks-part1.md
barriebyron, line 13:
```suggestion
Having said that, there are important reasons to keep some tests running on Mina's PRs: For example, many cryptography changes have the potential to break proof generation on SnarkyJS, and by far the best place to address that is the PR introducing the cryptography change.
```

0003-minimal-snarkyjs-checks-part1.md
```diff
@@ -0,0 +1,91 @@
...
+
+Having said that, there are important reasons to keep some tests running on Mina's PRs: For example, many cryptography changes have the potential to cause proof generation on SnarkyJS, and by far the best place to address that is the PR introducing the cryptography change.
+
+As of this RFC, all the SnarkyJS are configured to "soft-fail" on Mina's CI; this was a quickfix to empower quick Mina development, but is creating opportunities for bitrot in these situations so this RFC aims to quickly reconcile this temporary situation.
```
barriebyron, line 15: ummmm I am not sure what the subject of "all the SnarkyJS" refers to 

(updates?)

0003-minimal-snarkyjs-checks-part1.md
```diff
@@ -0,0 +1,91 @@
...
+
+Success Metric: Protocol developers are undisturbed by SnarkyJS tests in day-to-day development, but when deep changes to cryptography are taken,
```
barriebyron, line 17:
```suggestion
Success Metric: Protocol developers are undisturbed by SnarkyJS tests in day-to-day development. However, when deep changes to cryptography are made, the following requirements apply.
```

barriebyron: something like this to introduce the requirements (I think we mean the requirements apply only to deep changes)

0003-minimal-snarkyjs-checks-part1.md
```diff
@@ -0,0 +1,91 @@
...
+
+Requirements:
+
+* The bare-bones SnarkyJS tests (as described below) run as blocking jobs on Mina PRs
+* Any SnarkyJS tests that are more tightly coupled to zkApps do not block Mina CI
+
+Non-requirements (future work):
+
+* The other SnarkyJS tests are executing automatically by CI somewhere (probably the SnarkyJS repo)
+* The submodule structure is changed -- this should be done carefully in conjunction with moving Pickles and other cryptography layers out of Mina; perhaps we can remove safely remove it altogether
```
barriebyron, line 27:
```suggestion
* The submodule structure is changed -- this structure change must be done carefully in conjunction with moving Pickles and other cryptography layers out of Mina; perhaps we can remove safely remove it altogether
```

barriebyron (commented):
just a few questions and suggestions to improve readability

0003-minimal-snarkyjs-checks-part1.md
```diff
@@ -0,0 +1,91 @@
+## Summary
```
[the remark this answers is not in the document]
barriebyron: As a reader of a document, I like to have the title more recognizable than "Summary" wdyt?

0003-minimal-snarkyjs-checks-part1.md
```diff
@@ -0,0 +1,91 @@
...
+[detailed-design]: #detailed-design
+
+To meet the requirements, it suffices to extract a single creation of an end-to-end recursive proof in type-free SnarkyJS into a separate CI job:
+
+(TODO bkase: Move this to a comment in the test, and link to it in the RFC)
+* This test will compile and create a recursive proof of one layer using ZkProgram
+* This test will run with type-checking turned off on TypeScript, this way we won't worry about type schema changes
+* Since the smart contract interface isn't used, we don't need to worry about anything breaking if transaction or account structure changes
+
+The existing SnarkyJS will remain unchanged in the "soft-fail" job on the Mina repo.
```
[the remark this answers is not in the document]
mitschabaude: I don't think so! It depends on what workflow we envision:

a) We envision a workflow where the protocol team makes breaking changes without being blocked, but SnarkyJS developers are still _notified_ of a breaking change and will try to fix it within the next week or so. Advantage: We catch breakage as it happens, and always know what PR caused it. In this scenario, the soft-fail test is the necessary signal to notice when to fix something.

b) SnarkyJS development doesn't try to keep up with protocol's changes at all. We only fix up SnarkyJS before a new protocol version is "released" (say, we deploy a new testnet). That implies the soft-fail test in Mina CI will permanently be broken and won't be a useful signal, and you're right that the test is not useful in that case.

I strongly prefer a) over having a huge diff of commits from dozens of PRs and trying to figure out what in that list might have broken something.

I think the relation of submodules is not tied to what checks we want to have in which repo's CI. In snarkyjs CI, we rely on prebuilt bindings artifacts (we don't compile Rust and OCaml from source). Whenever you create new bindings artifacts, you also update a file called `MINA_COMMIT` which logs the Mina commit that the bindings were created from. So, we already have that system you describe for being able to see Mina changes transparently within the snarkyjs history, without relying on git submodules.

0003-minimal-snarkyjs-checks-part1.md
mitschabaude: I'd say about half, or slightly more, of the existing unit tests don't depend on Mina types. So yeah, we could compile a subset of them from TS to JS without type-checking, and run them as part of the new blocking CI job. We'll have to see what tests exactly we want there.

I think what Brandon describes is the most minimal version that still would cover almost all breaking crypto changes

> * This test will compile and create a recursive proof of one layer using ZkProgram
> * This test will run with type-checking turned off on TypeScript, this way we won't worry about type schema changes

mitschabaude (commented):
I think this RFC describes the correct first step of redesigning the snarkyjs - mina code relationship!

mitschabaude:
I'll comment on @nholland94 question:

> To help me understand the reason behind the approach, can we expand on why it makes sense to continue to keep the tests inside of the Mina repository? My question is 2-fold:
> 
> 1. If we are just keeping the SnarkyJS test suite as a soft fail in the repo, isn't this strictly more difficult to deal with than inverting the submodules? Eg: The Mina protocol code and SnarkyJS code is out of date -- if I have Mina as a submodule to SnarkyJS, then I can diff the changes that caused the mismatches when I updated the submodule; if I have things as is, I have to go through the commit history, look at CI, see where things started failing, and then work through that diff.

Answered this here: https://github.com/o1-labs/rfcs/pull/5#discussion_r1227572677, I think soft-fail test is what we want regardless of submodule relationship.

The principle that motivated the old architecture of snarkyjs tests, and still upheld with this RFC, is the following:

**Tests should reveal breaking changes immediately within the same PR that is making the change**

Having this makes it much easier to fix something, in my experience.

> 
>  2. What class of issues does the end-to-end recursive proof test prevent compared to the full SnarkyJS smart contract test suite?

The goal of a full end to end recursive proof test would be to catch all failures in the SNARK cycle (compile / prove / verify) which are introduced by crypto changes. It's very easy to introduce such failures, for example, just by changing one of the Rust-to-OCaml APIs and not updating the untyped wasm-to-jsoo JS translation file. These tend to be hard to debug and fix if not coupled to the PR that introduced them.

> Additionally, this document does not currently say anything about the Mina integration tests which rely on SnarkyJS. If those are still in and blocking, then schema changes will still break Mina's CI, since we are actually executing the smart contract code of SnarkyJS in those tests.

Good call-out! The snarkyjs-mina integration tests necessarily depend on protocol so they should be treated the same as the currently disabled full snarkyjs unit tests - they should be soft-failing to notify us of issues, without blocking.

As an aside, this test is pretty important as it's currently the only place where snarkyjs interaction with a MIna node, and even just graphql fetching in snarkyjs, are tested. But @shimkiv is working on adding similar tests to snarkyjs CI.

mitschabaude (approved):

0003-minimal-snarkyjs-checks-part1.md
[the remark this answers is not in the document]
barriebyron: template now had a heading level 1 to identify the RFC by name, UX change for readability and consumability

nholland94:
I still don't believe this makes product engineering lives better, but this plan certainly improves things for protocol engineers. So long as product is happy, feel free to merge.

mrmr1993 (approved):

nholland94 (approved):

Issue #7, opened 2023-06-14
Draft RFC for Dockerised LocalNetwork


nicc:
It should follow the template provide.
It should describe usage of the new local network option.
It should describe downstream impact and opportunities - any tools and/or infrastructure that can or should be updated.

shimkiv:
https://github.com/o1-labs/rfcs/pull/8

PR #8, opened 2023-06-14
[RFC] Mina Local Network as Docker image

shimkiv:
Closes https://github.com/o1-labs/rfcs/issues/7

---

[Internal discussion](https://o1-labs.slack.com/archives/C028Q27R8UC/p1686392840257539).


0004-dockerised-local-network.md
```diff
@@ -0,0 +1,68 @@
...
+
+The Dockerization of lightweight Mina Local Networks aims to address these challenges by providing a lightweight, fast, and resource-efficient solution for spinning up local Mina networks. This will significantly improve the DX/UX, allowing developers to easily test their applications and changes against a "close to real" Mina network.
+
+## Detailed design
+
+The Dockerization process involves two main phases:
+
+1. **Building Phase**: Usual build procedure the engineers familiar with. In this phase, the Mina Daemon is built locally or on a CI server using the desired branch, revision, and compile-time constants. The build process can leverage the `lightnet` Dune profile to create a lightweight version of the Mina Daemon.
+2. **Dockerization Phase**: In this phase, scripts are invoked to prepare the filesystem (binaries, additional scripts, genesis ledger configuration with pre-funded accounts, etc.) and ensure that the Mina Local Network scripts work well inside the Docker environment. The Docker image is then built and uploaded to Docker Hub using provided credentials (Users have to be authenticated using `docker login` CLI).
```
barriebyron, line 18:
```suggestion
2. **Dockerization Phase**: In this phase, scripts are invoked to prepare the file system (binaries, additional scripts, genesis ledger configuration with pre-funded accounts, etc.) and ensure that the Mina Local Network scripts work well inside the Docker environment. The Docker image is then built and uploaded to Docker Hub using provided credentials. (Users have to be authenticated using [`docker login`](https://docs.docker.com/engine/reference/commandline/login/) CLI).
```

0004-dockerised-local-network.md
```diff
@@ -0,0 +1,68 @@
...
+3. **Testing scope**: The testing will cover all major functionalities of the Docker image, including Dockerization phase, and the resulting Mina Local Network operation.
+4. **Testing requirements**: The testing will require a suitable environment for running Docker, as well as tools for monitoring resource usage and blockchain behavior.
+5. **Testing resources**: The testing will be conducted on a local machine. The testing may also require additional tools for monitoring and debugging.
+
+## Drawbacks
+
+One drawback of this approach is the potential for discrepancies between the behavior of the lightweight Mina Local Network and the full Mina network. While the lightweight network is designed to be "close to real", it may not perfectly replicate all aspects of the full network, which could lead to misleading test results. To mitigate this drawback, we can state clearly in the documentation that the lightweight network is not intended to be a perfect replica of the full network and that users/developers should always double check their work against real networks.
```
barriebyron, line 49:
```suggestion
One drawback of this approach is the potential for discrepancies between the behavior of the lightweight Mina Local Network and the full Mina network. While the lightweight network is designed to be "close to real", it may not perfectly replicate all aspects of the full network, which could lead to misleading test results. To mitigate this drawback, we can state clearly in the documentation that the lightweight network is for local development only and is not intended to be a perfect replica of the full network. Mina Local Networks do not guarantee production-ready results. Users and developers must always double-check their work against real networks.
```

barriebyron (commented):
very thorough @shimkiv with excellent substantiation and details

I'll let the required architects approve the RFC proper.
This comment approves the editorial content and well-described narrative.

0004-dockerised-local-network.md
shimkiv: Thank you!

0004-dockerised-local-network.md
shimkiv: Thank you, and I'll change the case to lower.

0004-dockerised-local-network.md
```diff
@@ -0,0 +1,68 @@
...
+
+The Mina protocol, while powerful and innovative, currently lacks a simple and efficient way to deploy and run local networks for testing purposes. This has led to challenges in both development and user experience (DX/UX), as developers have to either use costly cloud solutions or set up everything locally, both of which can be slow and resource-intensive.
+
+The Dockerization of lightweight Mina Local Networks aims to address these challenges by providing a lightweight, fast, and resource-efficient solution for spinning up local Mina networks. This will significantly improve the DX/UX, allowing developers to easily test their applications and changes against a "close to real" Mina network.
+
+## Detailed design
+
+The Dockerization process involves two main phases:
+
+1. **Building Phase**: Usual build procedure the engineers familiar with. In this phase, the Mina Daemon is built locally or on a CI server using the desired branch, revision, and compile-time constants. The build process can leverage the `lightnet` Dune profile to create a lightweight version of the Mina Daemon.
```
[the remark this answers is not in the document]
shimkiv: Thank you, added the link to new compile time Mina Daemon constants configuration.

0004-dockerised-local-network.md
```diff
@@ -0,0 +1,68 @@
...
+
+Another potential drawback is the need for Docker as a dependency. Users who do not have Docker installed or who are unfamiliar with Docker may face additional hurdles in setting up and using the Mina local network.
+
+## Rationale and alternatives
+
+The Dockerization of Mina local networks is proposed as a solution to the current challenges in deploying and running local Mina networks for testing purposes. By encapsulating the Mina daemon and other necessary components in a Docker image, we can provide a simple, fast, and resource-efficient solution for spinning up local Mina networks.
+
+An alternative to this approach would be to improve the existing methods for deploying and running Mina networks, such as optimizing the setup process or providing better support for cloud-based solutions. However, these alternatives would likely be more complex and time-consuming to implement and would not offer the same level of simplicity and efficiency as the Docker-based solution.
```
mitschabaude, line 57: Viable Systems develop the [OpenMina](https://openmina.com/) node, which targets web / JS environments.

This goes a bit beyond this RFC, but I think it would be an interesting alternative to use OpenMina instead of the OCaml Mina node, especially for when we want to package the local Mina node as part of JS tooling like the zkapp CLI.

mitschabaude (commented):
Thanks for putting this up as an RFC!

Since a major use case for this is testing of client-side tools like snarkyjs and the zkapp-CLI, I'd like to see the specific needs of those tools addressed in more detail.

> The Docker image can also be used in CI jobs to run end-to-end tests, particularly for tools that don't require fully fledged networks but still need to test integration with "close to real" networks.

I think that "close to real" networks doesn't precisely capture what we want for testing client-side tools. (We do need close to real networks for e2e tests, though.)

What we want in CI for client-side tools is a way to _run transactions against a local Graphql endpoint_. That endpoint has to faithfully emulate the behaviour of a real Mina node, from the piont of view of a client.

What we _don't_ need in those tests is a close to real behavior in the internal operations of a node, or in the interactions between multiple nodes. In particular, I argue that we're not interested in:
- networking
- consensus
- startup phase of a Mina node
- snark workers / blockchain proofs (as you mentioned)

Anything included in the docker image that uses resources (RAM usage etc) in order to test one of those would be overkill, and should be avoided.

So, my question would be: Can we have a version of this docker image that
- avoids running multiple nodes altogether?
  - but still produces blocks?
- has no expensive startup phase?
- uses less than 2GB of RAM?

0004-dockerised-local-network.md
shimkiv: @mitschabaude thank you for review and comments!

> avoids running multiple nodes altogether?
> but still produces blocks?

Yes we can have it, new implementation (or packaging) runs `one-node-network` but still behaves like complete blockchain.

> has no expensive startup phase?

Startup time dropped from 4+ minutes down to 1-1.5 minutes.

> uses less than 2GB of RAM?

Consumes `~625-650Mb` of RAM after initial startup spike.

> This goes a bit beyond this RFC, but I think it would be an interesting alternative to use OpenMina instead of the OCaml Mina node, especially for when we want to package the local Mina node as part of JS tooling like the zkapp CLI.

It will be definitely good to see what we can get from `OpenMina`.

0004-dockerised-local-network.md
```diff
@@ -0,0 +1,77 @@
...
+## Prior art
+
+Dockerization is a common practice in software development, particularly for applications that require complex setup processes or that need to run in isolated environments. Many blockchain projects also provide Docker images for their nodes or other components, recognizing the benefits of Docker in terms of simplicity, reproducibility, and resource efficiency.
+
+## Unresolved questions
+
+- What is the best way to handle the building phase? And how can we ensure that the Docker image is kept up-to-date with the latest changes to the Mina Daemon and other components?
+  - Should we add additional Minaprotocol CI job to periodically (say, once a day) build the lightweight Mina Daemon, build Docker image and publish it on behalf of O(1) Labs Docker Hub account?
+  - Which Mina branches should be used to prepare the corresponding Docker images with the lightweight Mina Network inside?
+- How can we best handle the potential discrepancies between the lightweight Mina local network and the full Mina network?
```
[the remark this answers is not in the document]
shimkiv: https://github.com/o1-labs/rfcs/pull/8/commits/8f8624e3fadc06ee4932ae55ecfd51d08d2f5341

0004-dockerised-local-network.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Detailed design
+
+The Dockerization process involves two main phases:
+
+1. **Building Phase**: Usual build procedure the engineers familiar with. In this phase, the Mina Daemon is built locally or on a CI server using the desired branch, revision, and compile-time constants. The build process can leverage the [lightnet](https://github.com/MinaProtocol/mina/blob/4e0b324912017c3ff576704ee397ade3d9bda412/src/config/lightnet.mlh) Dune profile to create a lightweight version of the Mina Daemon.
+2. **Dockerization Phase**: In this phase, scripts are invoked to prepare the file system (binaries, additional scripts, genesis ledger configuration with pre-funded accounts, etc.) and ensure that the Mina local network scripts work well inside the Docker environment. The Docker image is then built and uploaded to Docker Hub using provided credentials. (Users have to be authenticated using [`docker login`](https://docs.docker.com/engine/reference/commandline/login/) CLI).
```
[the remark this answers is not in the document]
shimkiv: https://github.com/o1-labs/rfcs/pull/8/commits/8f8624e3fadc06ee4932ae55ecfd51d08d2f5341

0004-dockerised-local-network.md
```diff
@@ -0,0 +1,77 @@
...
+Another future option might be in utilizing [OpenMina](https://openmina.com/) project, which can target `Web` / `JS` runtime and provide a lightweight Mina local network in a browser. This would be a great option for developers, however, it is not available at the moment.
+
+## Prior art
+
+Dockerization is a common practice in software development, particularly for applications that require complex setup processes or that need to run in isolated environments. Many blockchain projects also provide Docker images for their nodes or other components, recognizing the benefits of Docker in terms of simplicity, reproducibility, and resource efficiency.
+
+## Unresolved questions
+
+- What is the best way to handle the building phase? And how can we ensure that the Docker image is kept up-to-date with the latest changes to the Mina Daemon and other components?
+  - Should we add additional Minaprotocol CI job to periodically (say, once a day) build the lightweight Mina Daemon, build Docker image and publish it on behalf of O(1) Labs Docker Hub account?
```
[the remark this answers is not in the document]
shimkiv: I don't think we need to build it on every PR or on every push/merge into main branches since doing so will either overwrite existing images too often or we will end up storing good amount of "legacy" images, in which case the clean-up job will be required.
In addition we don't know the actual demand, what people really expect of it.
Taking above into consideration I'd start simple and just do Nightly builds for now.

0004-dockerised-local-network.md
```diff
@@ -0,0 +1,77 @@
...
+  - Which Mina branches should be used to prepare the corresponding Docker images with the lightweight Mina Network inside?
```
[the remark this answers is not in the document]
shimkiv: Similarly, I'd start with main branches only for now. 
With, perhaps, an ability for anyone to build their own images or at least documenting the process.

bkase (approved):

PR #9, opened 2023-06-15
SnarkyJS SHA/Keccak

Trivo25:
ECDSA in separate PR https://github.com/o1-labs/rfcs/pull/14, based on this one.

### DIFF
--- a/0005-snarkyjs-ecdsa-sha.md
+++ b/0005-snarkyjs-ecdsa-sha.md
@@ -77,28 +77,9 @@ const Sha3_512 = buildSha(512, true);
 const Keccak = buildSha(256, false);
 ```
 
-Another alternative to the factory pattern above could be to supply the developer with a single function that takes a range of parameters, so that the developer can choose their flavour of SHA3/Keccak on their own.
+### Implementation (previously Alternative Approach)
 
-_Note_: Currently, if `nist` is set to `false`, we only support an output length of 256.
-
-```ts
-function SHA3(length: 224 | 256 | 385 | 512 | { nist: 256 }): {
-  hash(xs: UInt8[]);
-};
-```
-
-Developers can then import these functions into their project via
-
-```ts
-import { Sha3_224, Sha3_256, Sha3_385, Sha3_512, Keccak } from "snarkyjs";
-
-Sha3_224.hash(xs);
-// ..
-```
-
-### Alternative Approach
-
-Another possibility is to combine all hash functions, including Poseidon, under a shared namespace `Hash`. Developers will then be able to use these functions by calling `Hash.[hash_name].hash(xs)`. However, this would not be equivalent to the existing `Poseidon` API.
+After discussing different API approaches, the overall consensus agreed on pursuing this particular approach over others. The implementation will combine all hash functions, including Poseidon, under a shared namespace `Hash`. Developers will then be able to use these functions by calling `Hash.[hash_name].hash(xs)`. However, this would not be equivalent to the existing `Poseidon` API.
 
 However, the existing `Poseidon` API could be deprecated in favour of the new `Hash` namespace.
 
@@ -131,6 +112,27 @@ Hash.Keccak.hash(xs);
 
 // deprecated in favor of Hash.Poseidon.hash
 Poseidon.hash(xs);
+```
+
+### Alternative Approach
+
+Another alternative to the factory pattern above could be to supply the developer with a single function that takes a range of parameters, so that the developer can choose their flavour of SHA3/Keccak on their own.
+
+_Note_: Currently, if `nist` is set to `false`, we only support an output length of 256.
+
+```ts
+function SHA3(length: 224 | 256 | 385 | 512 | { nist: 256 }): {
+  hash(xs: UInt8[]);
+};
+```
+
+Developers can then import these functions into their project via
+
+```ts
+import { Sha3_224, Sha3_256, Sha3_385, Sha3_512, Keccak } from "snarkyjs";
+
+Sha3_224.hash(xs);
+// ..
 ```
 
 It is important to mention that we have a lot of freedom in the the implementation details of the API, as the gates and gadgets are already implemented in OCaml and just need to exposed and wrapped by a developer friendly API.


0005-snarkyjs-ecdsa-sha copy.md
```diff
@@ -0,0 +1,132 @@
...
+    },
+  };
+}
+
+const Sha3_224 = buildSha(224, true);
+const Sha3_256 = buildSha(256, true);
+const Sha3_385 = buildSha(385, true);
+const Sha3_512 = buildSha(512, true);
+const Keccak = buildSha(256, false);
+```
```
Trivo25, line 86: Another API could be to provide the developers with a function `function SHA3(length: 224 | 256 | 385 | 512): { hash(xs: Field[]) }`, allowing the developer to "build" their SHA variant on-demand, while providing Keccak separately

0005-snarkyjs-ecdsa-sha copy.md
```diff
@@ -0,0 +1,132 @@
...
+[Keccak PRD](https://www.notion.so/minaprotocol/Keccak-gadget-PoC-PRD-59b024bce9d5441c8a00a0fcc9b356ae) then go into detail and describe two of the most important building blocks to achieve Ethereum compatibility in a cryptographic sense.
+
+This RFC describes the steps needed to expose both ECDSA and SHA3/Keccak to SnarkyJS, enabling SnarkyJS developer to use these primitives in their applications and smart contracts.
+
+It is important to mention that this RFC only considers the work required to _expose_ already existing cryptographic primitives (gates and gates), which have previously been implemented by the crypto team, from the OCaml bindings layer to SnarkyJS - it is not required to implement additional cryptographic primitives. Changes will only impact [snarkyjs-bindings](https://github.com/o1-labs/snarkyjs-bindings) and [SnarkyJS](https://github.com/o1-labs/snarkyjs) itself.
```
nicc, line 22: Should this be "gates and gadgets"?

0005-snarkyjs-ecdsa-sha copy.md
```diff
@@ -0,0 +1,132 @@
...
+
+## ECDSA
+
+**TODO** this will be a seperate PR stacked on top of this one
+
+**Evergreen, wide-sweeping Details**
+
+**Ephemeral details**
+
+## Test plan and functional requirements
```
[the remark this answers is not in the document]
Trivo25: uupps

0005-snarkyjs-ecdsa-sha.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Detailed design
+
+[detailed-design]: #detailed-design
+
+### SHA3/Keccak
+
+The Keccak and SHA3 gadget has been implemented by the crypto team, enabling us to leverage the already existing bindings layer to and from OCaml. This design allows us to simply integrate the new gadgets into SnarkyJS by simply exposing them in `ocaml/lib/snarky_js_bindings_lib.ml`.
+
+For Keccak/SHA3, the implementation exposes two ready-to-go functions.
```
MartinMinkov, line 34: Nit: It'd be nice to have a mention to a specific commit/line of the implementation in the callout

0005-snarkyjs-ecdsa-sha.md
```diff
@@ -0,0 +1,197 @@
...
+In order to test the implementation of SHA3 and Keccak in SnarkyJS, we will follow the testing approach we already apply to other gadgets and gates.
+This includes testing the out-of- and in-snark variants using our testing framework, as well as adding a SHA3 and Keccak regression test. The regression tests will also include a set of predetermined digests to make sure that the algorithm doesn't unexpectedly change over time (similar to the tests implemented for the OCaml gadget).
+
+In addition to that, we should provide a dedicated integration test that handles SHA3/Keccak hashing within a smart contract (proving enabled). This will allow us to not only provide developers with an example, but also ensure that SHA3 and Keccak proofs can be generated.
+
+## Drawbacks
+
+[drawbacks]: #drawbacks
+
+Compared to Poseidon, hashing with SHA3 and Keccak is expensive. This should be made clear to the developer to avoid inefficient circuits. Additionally, it is important to educate developers of when to use SHA3/Keccak and when to use Poseidon. Additionally, the API should be secure.
```
MartinMinkov, line 178: When should developers use `SHA/Keccack` vs `Poseidon` It would be helpful to have context in the RFC

0005-snarkyjs-ecdsa-sha.md
```diff
@@ -0,0 +1,197 @@
...
+Developers can then import these functions into their project via
+
+```ts
+import { Sha3_224, Sha3_256, Sha3_385, Sha3_512, Keccak } from "snarkyjs";
+
+Sha3_224.hash(xs);
+// ..
+```
+
+### Alternative Approach
```
MartinMinkov, line 106: I've spent some time researching and thinking about our approach to the API, and I feel a Hash interface could be a great fit for us.

I took a look into the functionality of various popular JS runtimes like Node.JS, Deno, and Bun, I noticed they offer builder patterns for creating hash functions. While their approach is great, considering the diverse configuration options they have to deal with, our needs aren't as complex (I think). So, perhaps we might not need to follow the builder pattern for our hash functions.

I also looked into how Noir structures its approach: https://noir-lang.org/standard_library/cryptographic_primitives/hashes. 

They follow a `std:hash:[hash_name]` style, influenced by Rust, which I found quite appealing. It feels efficient and neat to call the function directly without any additional building process. My vote goes towards an interface-styled approach, happy to hear other opinions, though.

0005-snarkyjs-ecdsa-sha.md
```diff
@@ -0,0 +1,197 @@
...
+
+Another possibility is to combine all hash functions, including Poseidon, under a shared namespace `Hash`. Developers will then be able to use these functions by calling `Hash.[hash_name].hash(xs)`. However, this would not be equivalent to the existing `Poseidon` API.
+
+However, the existing `Poseidon` API could be deprecated in favour of the new `Hash` namespace.
```
mitschabaude, line 110: Given that others prefer this approach, I'm ok with moving `Poseidon` to a new `Hash` namespace. I agree with Brandon that this probably makes it easier to find.

The concern I have is that such an interface does not single out Poseidon among all the hash functions. To the contrary, since `SHA3` and `Keccak` are more familiar, non-expert developers could default to one of these instead of Poseidon. However, Poseidon as the hash function that work on native field elements, and which our proof system is specifically optimized for, should clearly be the default choice. It takes 11 constraints where Keccak takes many thousands.

To single out Poseidon as _the_ recommended hash function, one approach could be to also add a top level `hash()` method:

```ts
import { hash, Hash } from "snarkyjs"

// the default hash function (= Poseidon)
hash([x,y,z]);

// other hash function for special needs and more advanced users
Hash.Keccak.hash([x,y,z]);
```

This way, we get a clear default hash function while removing the "Poseion" jargon (we can still explain what hash function this is in the doccomment)

0005-snarkyjs-ecdsa-sha.md
```diff
@@ -0,0 +1,197 @@
...
+```
+
+It is important to mention that we have a lot of freedom in the the implementation details of the API, as the gates and gadgets are already implemented in OCaml and just need to exposed and wrapped by a developer friendly API.
+
+One goal of the API is to design an easy to use interface for developers as well as maintaining consistency with already existing primitives (Poseidon in this case, Schnorr in the case of ECDSA).
+The usage of both Keccak/SHA3 and Poseidon is the same: The user passes an array of Field elements into the hash function and the output is returned. One important detail is that Keccak/SHA3 only accepts an array of Field element that each are no larger than 1 byte. We will also provide additional helper functions that allow conversion between Field element arrays and hexadecimal encoded strings.
+
+```ts
+function fieldBytesFromHex(xs: Field[]): string;
+function hexToFieldBytes(hex: string): Field[];
```
mitschabaude, line 150: thanks for calling it out that the Keccak interface expects single-byte field elements.

I think we should follow our usual approach here, and introduce a provable type which encodes the constraint of being no more than a single byte long: `UInt8` or `Byte`

The Keccak / SHA3 interface should therefore be:
```ts
function hash(message: UInt8[]): UInt8[]
// TODO what is the return type? we could return a fixed length tuple!
```

mitschabaude: Btw @Trivo25, your RFC has the `fieldBytesFromHex` and `hexToFieldBytes` interfaces mixed up

0005-snarkyjs-ecdsa-sha.md
mitschabaude: Agree with having the different hashes directly!

nicc: +1

0005-snarkyjs-ecdsa-sha.md
nicc: +1

ymekuria: I like the idea of making `Poseidon` the default but having both a `hash` and `Hash` in SnarkyJS seems a little confusing although a developer can get more info with intellisense.

0005-snarkyjs-ecdsa-sha.md
ymekuria: +1 That will be very helpful.

barriebyron:
I removed the extraneous Markdown that breaks CI, the template was updated after this RFC was created https://github.com/o1-labs/rfcs/pull/10

0005-snarkyjs-ecdsa-sha.md
```diff
@@ -0,0 +1,179 @@
...
+**Evergreen, wide-sweeping Details**
+
+**Ephemeral details**
+
+## Test plan and functional requirements
+
+## SHA3/Keccak
+
+In order to test the implementation of SHA3 and Keccak in SnarkyJS, we will follow the testing approach we already apply to other gadgets and gates.
+This includes testing the out-of- and in-snark variants using our testing framework, as well as adding a SHA3 and Keccak regression test. The regression tests will also include a set of predetermined digests to make sure that the algorithm doesn't unexpectedly change over time (similar to the tests implemented for the OCaml gadget).
```
shimkiv, line 160: > a set of predetermined digests

If we're considering to utilize the test vectors from organizations like `NIST` (to ensure the correctness of the implementation), then we should specify it.

> regression test

It will be good to also mention what will form the regression testing suite (for example, functional tests, specific edge cases, errors handling, backward compatibility, etc.).

0005-snarkyjs-ecdsa-sha.md
```diff
@@ -0,0 +1,179 @@
...
+
+In addition to that, we should provide a dedicated integration test that handles SHA3/Keccak hashing within a smart contract (proving enabled). This will allow us to not only provide developers with an example, but also ensure that SHA3 and Keccak proofs can be generated.
```
shimkiv, line 162: - Are we concerned about performance testing? 
  - For example, hashing large amounts of data or rapidly making hash requests.
- Are we concerned about collision testing?
- Are we concerned about security testing? 
  - Possible/known vulnerabilities, cryptographic weakness, etc.
- Are we concerned about usability testing?
  - In a form of gathering the feedback from a small group of developers to improve the design and usability of the API before releasing it.

---

- The RFC indicates that hashing with SHA3 and Keccak is more expensive than using Poseidon. Performance metrics, such as time and memory usage, should be gathered and analyzed to provide developers with a clear understanding of the subject.

0005-snarkyjs-ecdsa-sha.md
mitschabaude: FYI, we have a "regression test suite" in use in snarkyjs CI, which this RFC refers to. it checks that constraints generated using a given list of contracts / provable methods doesn't change: https://github.com/o1-labs/snarkyjs/blob/main/src/examples/vk_regression.ts

0005-snarkyjs-ecdsa-sha.md
```diff
@@ -0,0 +1,179 @@
...
+```ocaml
+(** Gagdet for NIST SHA-3 function for output lengths 224/256/384/512 *)
+val nist_sha3 :
+    (module Snarky_backendless.Snark_intf.Run with type field = 'f)
+  -> int
+  -> 'f Snarky_backendless.Cvar.t list
+  -> 'f Snarky_backendless.Cvar.t array
+
+(** Gadget for Keccak hash function for the parameters used in Ethereum *)
+val eth_keccak :
```
querolita, line 35: Update function name to `ethereum`

0005-snarkyjs-ecdsa-sha.md
```diff
@@ -0,0 +1,179 @@
...
+These two functions will be imported into the bindings layer and exposed via a new sub-module as part of the already existing `Snarky` module.
+
+```ocaml
+
+  module Sha = struct
+    let create message nist length =
+      let message_array = Array.to_list message in
+      if Js.to_bool nist then
+        Kimchi_gadgets.Keccak.nist_sha3 (module Impl) length message_array
+      else Kimchi_gadgets.Keccak.eth_keccak (module Impl) message_array
```
querolita, line 51: Update here as well

0005-snarkyjs-ecdsa-sha.md
```diff
@@ -0,0 +1,179 @@
...
+```
+
+These two functions will be imported into the bindings layer and exposed via a new sub-module as part of the already existing `Snarky` module.
+
+```ocaml
+
+  module Sha = struct
+    let create message nist length =
+      let message_array = Array.to_list message in
+      if Js.to_bool nist then
```
querolita, line 49: Is `nist` meant to be the flag in the Snarky gadget? If so, after being asked in review of the Keccak PR, it got renamed into `nist_version`, still a boolean.

0005-snarkyjs-ecdsa-sha.md
```diff
@@ -0,0 +1,179 @@
...
+        Kimchi_gadgets.Keccak.nist_sha3 (module Impl) length message_array
+      else Kimchi_gadgets.Keccak.eth_keccak (module Impl) message_array
+  end
+
+```
+
+In order to reduce the bindings surface, Keccak and SHA3 will be exposed via a `create` function, which behaves like a factory pattern.
+By calling this function inside SnarkyJS, we can define and expose all possible variants of SHA3(224/256/384/512) and Keccak without the need to have an individual function in the bindings layer for each variant.
```
querolita, line 57: Note that the non-NIST variant is only implemented for 256-bit output. It could easily be extended to support all other variants as well, but since Ethereum uses the 256 one, and this is one of the very few applications where non-NIST version has been used, I preferred to simplify the options. Nonetheless, If the product team believe that it would be useful to have other variants, it is a one-liner update to the Snarky gadget. Let me know so I can act accordingly.

0005-snarkyjs-ecdsa-sha.md
```diff
@@ -0,0 +1,179 @@
...
+
+In order to differentiate Poseidon, which works over native Field elements, and SHA3 and Keccak which works over byte-sized Field elements, it will be beneficial to introduce a new type `UInt8` (a Field element that is exactly a byte) to draw a clan line between both hash functions.
```
querolita, line 59: Seems useful. Nonetheless, the gadget already performs all necessary constraints to make sure that the inputs are indeed 8 bits at most. That means it is sound. But if you want to provide the user interface with a much more explicit type for this so they can see if it is ok at compile time, then I believe it is a great idea for user experience.

0005-snarkyjs-ecdsa-sha.md
```diff
@@ -0,0 +1,179 @@
...
+const Sha3_512 = buildSha(512, true);
+const Keccak = buildSha(256, false);
+```
+
+Another alternative to the factory pattern above could be to supply the developer with a single function that takes a range of parameters, so that the developer can choose their flavour of SHA3/Keccak on their own.
+
+```ts
+function SHA3(
+  length: 224 | 256 | 385 | 512,
+  nist?: boolean = true
```
querolita, line 85: Yeah, perhaps you should state quite clearly in the docs or somewhere else that for now the `nist_version = false` only works if the provided length is 256. Meaning, even if a user wants `Keccak.ethereum` for output length 512, what it will obtain is 256 bit output. Again, let me know if product wants to extend this functionality because it should be straightforward to add.

0005-snarkyjs-ecdsa-sha.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Drawbacks
+
+Compared to Poseidon, hashing with SHA3 and Keccak is expensive. This should be made clear to the developer to avoid inefficient circuits. Additionally, it is important to educate developers of when to use SHA3/Keccak and when to use Poseidon. Additionally, the API should be secure.
+Adding new primitives, especially cryptographic primitives, always includes risks such as the possibility of not constraining the algorithm and input enough to provide the developer with a safe API that is required to build secure applications. However, adding these primitives to SnarkyJS enables developers to explore a new range of important use cases.
+
+## Rationale and alternatives
+
+Keccak and SHA3 could not be exposed to SnarkyJS at all. However, this would essentially render these primitives useless since they were specifically designed to be used by developers with SnarkyJS. By adding these primitives, SnarkyJS will become an even more powerful zero-knowledge SDK that enables developers to explore a wide range of use cases.
```
querolita, line 171: I believe we should indeed expose it since this is a PoC, and if they found a flaw we would definitely address it. If it lives behind the hood, then it is less likely that users could spot a bug or trace it down to Keccak (if at all, luckily none).

querolita (approved):
Approving for now but please take a look at the comments to update the names, and whether or not you want to extend non-NIST to support other output lengths as well.

0005-snarkyjs-ecdsa-sha.md
mitschabaude: This is an interesting discussions! If we had a `UInt8` type, it would actually be preferable to constrain its range _as it is introduced into provable code_. This is how we usually do it, so that any user of `UInt8` in provable code can rely on the fact that this value can only be 8 bits long. (You might have seen the same done for booleans in snarky).

So, could it be a possibility to add a flag to the existing gadgets that disables the 8 bit range check on inputs? @querolita

0005-snarkyjs-ecdsa-sha.md
MartinMinkov: We could have the `Poseidon` hash function set as the default hash on the `Hash` object/namespace, like this:

```ts
import { Hash } from "snarkyjs"
// It can be invoked like this
Hash.default([x,y,z]);
// Or can be this
Hash.hash([x,y,z]);
```

This way, there's no need to deal with multiple top-level imports for `hash` and `Hash`, which could potentially lead to confusion. Personally, I'm finding the idea of using `default` as the terminology is quite appealing, as it's clear and straightforward.

mitschabaude: or maybe just
```ts
Hash([x,y,z])
```
although I don't like upper case for a function that's not a constructor

ymekuria: I like this option over the other two. 

```ts
// Or can be this
Hash.hash([x,y,z]);
```

mitschabaude: I'm not a fan of `Hash.default()`, but I'm ok with `Hash.hash()` even though it's a bit redundant. Personally I still prefer `hash()`

0005-snarkyjs-ecdsa-sha.md
MartinMinkov: I'd like to ask if adding `Field` as an acceptable type in the array is helpful. When `hash` is called, we can check if the field is exactly a byte at runtime. That way, we can be more flexible with our types by using `(Field | UInt8)[]`.

Returning tuples would be neat here! I think you keep all the normal array methods, so the added type safety is helpful!

0005-snarkyjs-ecdsa-sha.md
```diff
@@ -0,0 +1,179 @@
+# Exposing ECDSA and SHA3/Keccak to SnarkyJS
+
+## Summary
+
+Foreign cryptography primitives such as ECDSA and SHA3 are widely used outside of Mina. For example, Ethereum uses ECDSA over secp256k1 for signatures - in order to "communicate" with the outside world and other blockchains, SnarkyJS (and, therefore Mina) needs to support these primitives as well. This RFC describes how we will leverage the custom gates implemented by the crypto team and expose them to SnarkyJS, making them accessible to smart contract developers.
```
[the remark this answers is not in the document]
jspada: In case it's helpful, the `Bitwise` and `Keccak` gadgets are not foreign field based, but `Ecdsa`, `Ec_group` and `Foreign_field` gadgets are.

0005-snarkyjs-ecdsa-sha.md
jspada: Great idea @mitschabaude - makes sense.

Note typo: `draw a clan line`

0005-snarkyjs-ecdsa-sha.md
jspada: Just checking... in the snark these hashes come with some circuit constants and it may be more efficient to include those in the circuit once at the start of the circuit.  If the "direct hash" API isn't able to facilitate this, then the "builder" approach could have an advantage since in that approach you could add those constraints when building the hash configuration.

0005-snarkyjs-ecdsa-sha.md
jspada: +1 for specific edge cases, errors handling

Since padding is important for security and there are at least two kinds of padding supported by our Keccak gadgets, from a security perspective testing different length messages (e.g. empty, zero, one less than padding lengths, one more, multiples of, etc) are great edge-cases to test imho!

0005-snarkyjs-ecdsa-sha.md
```diff
@@ -0,0 +1,179 @@
...
+Adding new primitives, especially cryptographic primitives, always includes risks such as the possibility of not constraining the algorithm and input enough to provide the developer with a safe API that is required to build secure applications. However, adding these primitives to SnarkyJS enables developers to explore a new range of important use cases.
+
+## Rationale and alternatives
+
+Keccak and SHA3 could not be exposed to SnarkyJS at all. However, this would essentially render these primitives useless since they were specifically designed to be used by developers with SnarkyJS. By adding these primitives, SnarkyJS will become an even more powerful zero-knowledge SDK that enables developers to explore a wide range of use cases.
+
+## Prior art
+
+Exposing gates and gadgets from the OCaml layer to SnarkyJS is nothing new - the same procedure has been applied to other primitives such as Poseidon, Field, and Elliptic Curve operations.
+
```
jspada, line 176: Something not covered here that definitely is something users will require...

There is a layer between Keccak and ECDSA (not part of Keccak and not part of ECDSA), which is a gadget to constrain the mapping out Keccak output to foreign field elements.

This gadget is called `Foreign_field.bytes_to_standard_element`.  It takes as arguments an array of byte, a modulus $m$, the bitlength of the modulus and constrains the mapping (and modular reduction) of the bytes to the foreign field element in $[0, m-1]$

jspada (approved):

0005-snarkyjs-ecdsa-sha.md
mitschabaude: Good point! I think we'll introduce a new class `ForeignField` or similar (draft version here: https://github.com/o1-labs/snarkyjs/pull/985). We could include the conversion gadget you mention as a method
```ts
ForeignField.fromBytes(bytes: UInt8[]): ForeignField;
```
(the modulus is part of the ForeignField class, which is used with a class factory that takes the modulus)

0005-snarkyjs-ecdsa-sha.md
querolita: Answering your question @mitschabaude, yes, it could be easily added so that you have access to it from SnarkyJS. From a cryptography point of view, I still have the feeling that I do prefer to leave those constraints in the circuit. Meaning, even if the JS code won't allow you to introduce anything larger than UInt8, it still relies on trusting that the types are correct rather than a mathematical proof that that is indeed the case. But if this is something that SnarkyJS does often, I guess it is fine if you can disable the check in case the user introduces a UInt8 as input.

mitschabaude: @querolita I may have been not clear enough -- what I'm proposing is to add a mathematical proof that inputs fit in 8 bits. It's just better to add that proof when a value is introduced, and not inside a specific gadget.

Assume that we have multiple gadgets which operate on UInt8. If every gadget adds constraints to check that the Uint8 is only 8 bits long, then we will add those gates multiple times (if we use multiple of these gadgets).

If, instead, we add the constraints whenever a Uint8 is introduced (with `exists`), and not in each gadget, then we will have those constraints only once.

In snarkyjs and in snarky, we already have a way to add constraints to the proof everytime a variable of a certain type is introduced with `exists`. In snarky, every `typ` can have a `check` method which contains those "type-defining" constraints.

`exists` calls `typ.check`:
https://github.com/o1-labs/snarky/blob/14f8e2ff981a9c9ea48c94b2cc1d8c161301537b/src/base/checked_runner.ml#L277

and types such as booleans include their defining constraints as part of `typ.check`:
https://github.com/o1-labs/snarky/blob/14f8e2ff981a9c9ea48c94b2cc1d8c161301537b/src/base/utils.ml#L278-L281

querolita:
Hey! Since @mitschabaude suggested to include a flag to enable/disable  8-bit checks depending on whether the user introduced a list of Field or of UInt8, I thought maybe this could as well be interesting for outside users.

Right now, all keccak implementations consider the first element of the list to be the most significant byte of the string. Similar for the output. I wonder if it could be interesting to add a flag to indicate whether the input/output is desired to be in big endian or in little endian, to be more transparent about this point. What do you think? @Trivo25 @jspada

0005-snarkyjs-ecdsa-sha.md
querolita: Ooooooh!!! Okay I get you now! Nice very nice indeed! Actually we had discussions early in the days of the design of Keccak where we were analysing the possibility to add a "Byte" type, together with existing ones like Boolean or Field. But we thought it might not be super urgent at the Snarky level, but it indeed looks interesting at the SnarkyJS level. 

I really like this idea now. And yes, I can create a flag to disable these checks inside the gadget, if they have already been constrained at UInt8 creation time. I will create a subsequent PR with this, together with other improvements that have been suggested along this discussion.

0005-snarkyjs-ecdsa-sha.md
Trivo25: How about `Hash.digest`? It breaks the `.hash` pattern but also signals uniqueness

0005-snarkyjs-ecdsa-sha.md
Trivo25: Thanks everyone! I slightly edited this section to reflect your ideas. I am a big fan of adding (maybe not all, but at least some) of the NIST test vectors - what are your thoughts on this? Is this already being done in the crypto side? (didn't explicitly see that)

0005-snarkyjs-ecdsa-sha.md
mitschabaude: @jspada, don't we have a way to include constants only once anyway? "Constant unification"? When you use a constant that already exists in the circuit, it will just be wired to the existing one?
If that's true and we can use it, it seems to me we don't need a special setup command to avoid adding the same constants multiple times. cc @mrmr1993?

0005-snarkyjs-ecdsa-sha.md
bkase: I agree it's nice to have a default hash -- I also prefer `Hash.hash()` over `Hash.default()`. I'm mixed on the naked `hash()`.

I think we won't reach consensus so we need someone to make a decision. I think @nicc should decide these sorts of things until we have a PM on the team.

0005-snarkyjs-ecdsa-sha.md
bkase: +1 to removing the constraint checks in the gadgets and forcing input to be `UInt8` 👍

0005-snarkyjs-ecdsa-sha.md
bkase: In this case, assuming we don't need the other NIST lengths, we should definitely change the parameters of the function so that only the valid inputs typecheck.

Something like:

```typescript
function SHA3(length: 224 | 256 | 385 | 512 | { nist: 256 }, /*...*)
```

0005-snarkyjs-ecdsa-sha.md
Trivo25: Good question - @querolita to what extend has the original crypto implementation be tested? Do you have any recommendations to what we should test, if at all?

querolita: The tests performed to the implementation were:
- Checking different output lengths and different configurations of the padding (NIST's for SHA3 and submission's for Keccak)
- Checking different length outputs with different numbers of blocks
- Edge cases of padding to make sure that a whole new block is created, and messages that already look padded
- Checking different endianness of inputs and outputs
- Check that non-byte length inputs cannot be passed
- Can reuse constraint system for the same circuit structure
- Cannot reuse constraint system for different output lengths
- Cannot reuse constraint system for same output length but different padding
- Cannot reuse constraint system for different endianness

I would create similar tests for SnarkyJS, but probably more high level since you don't have access to low level details such as constraint system creation (that I am aware of). In particular, I would make sure that inputs which are not "correct formatted hex values" (meaning an odd number of hex digits) is not passed to the function. And of course making sure that all combinations actually produce the expected value (pairs of output lengths and padding types), for some random values. I wouldn't go much deeper than that regarding the snarkyjs level alone.

In that sense, we are not testing against collision, just comparing against official outputs of the hash function on those given inputs, nor performance testing (all we know is 1 block of encryption, which is enough for 1080bits of input data, fits in 15k kimchi rows, thus 4 full blocks can be hashed without chunking), we haven't looked for keccak's potential vulnerabilities, and we have not gathered usability info from users other than ourselves.

0005-snarkyjs-ecdsa-sha.md
```diff
@@ -0,0 +1,196 @@
...
+In order to reduce the bindings surface, Keccak and SHA3 will be exposed via a `create` function, which behaves like a factory pattern.
+By calling this function inside SnarkyJS, we can define and expose all possible variants of SHA3(224/256/384/512) and Keccak without the need to have an individual function in the bindings layer for each variant.
+
+In order to differentiate Poseidon, which works over native Field elements, and SHA3 and Keccak which works over byte-sized Field elements, it will be beneficial to introduce a new type `UInt8` (a Field element that is exactly a byte) to draw a clean line between both hash functions.
+
+In SnarkyJS, these new primitives will be declared as followed:
+
+```ts
+function buildSha(length: 224 | 256 | 385 | 512, nist: boolean) {
```
bkase, line 64: Can you update this with the most object too?

0005-snarkyjs-ecdsa-sha.md
```diff
@@ -0,0 +1,196 @@
...
+```ts
+import { Sha3_224, Sha3_256, Sha3_385, Sha3_512, Keccak } from "snarkyjs";
+
+Sha3_224.hash(xs);
+// ..
+```
+
+### Alternative Approach
+
+Another possibility is to combine all hash functions, including Poseidon, under a shared namespace `Hash`. Developers will then be able to use these functions by calling `Hash.[hash_name].hash(xs)`. However, this would not be equivalent to the existing `Poseidon` API.
```
bkase, line 101: From the conversation here, it seems like this is the approach we’re going with. Can you make the doc clear that this is the approach specified in the rfc and that the other approaches are the alternatives?

0005-snarkyjs-ecdsa-sha.md
bkase: @shimkiv does this sounds good to you? Is there anything else you would change here?

0005-snarkyjs-ecdsa-sha.md
Trivo25: What exactly do you mean by "the most object"?

0005-snarkyjs-ecdsa-sha.md
```diff
@@ -0,0 +1,198 @@
...
+
+```ts
+ForeignField.fromBytes(bytes: UInt8[]): ForeignField;
+```
+
+to the `ForeignField` implementation that constraints the output of Keccak to a foreign field, which will later be an important constraint for ECDSA.
+
+Overall, exposing new gadgets and gates follow a strict pattern that has been used all over in the SnarkyJS bindings layer. As an example, the [Poseidon](https://github.com/o1-labs/snarkyjs-bindings/blob/main/ocaml/lib/snarky_js_bindings_lib.ml#L386) implementation behaves similarly. From the point of view of SnarkyJS, these gadgets are just another set of function calls to OCaml.
+
+## ECDSA
```
shimkiv, line 164: Can we remove this section if there will be separate RFC?

0005-snarkyjs-ecdsa-sha.md
```diff
@@ -0,0 +1,198 @@
...
+
+**TODO** this will be a seperate PR stacked on top of this one
+
+**Evergreen, wide-sweeping Details**
+
+**Ephemeral details**
+
+## Test plan and functional requirements
+
+## SHA3/Keccak
```
shimkiv, line 174: `Test plan and functional requirements` and `SHA3/Keccak` sections are of the same level. Later one probably should be a sub-section (###).

0005-snarkyjs-ecdsa-sha.md
Trivo25: Yes! It will be removed with the other PR, which is stacked on top of this PR https://github.com/o1-labs/rfcs/pull/14

0005-snarkyjs-ecdsa-sha.md
shimkiv: > we will follow the testing approach we already apply to other gadgets and gates

It will be good to add some implementation example links here.

> will also include a set of predetermined digests

Can we please be more specific here? What kind of set? Is it standardised set? If not then how we determine the set to be good enough?

> similar to the tests implemented for the OCaml gadget

Can we please include links here?

> We will include a range of edge cases in the tests (e.g. empty input, zero, etc).

It will be good if we can be more explicit here and at least highlight the major cases we're aware of and that will be included into the testing plan.

> In that sense, we are not testing against collision, just comparing against official outputs of the hash function on those given inputs, nor performance testing (all we know is 1 block of encryption, which is enough for 1080bits of input data, fits in 15k kimchi rows, thus 4 full blocks can be hashed without chunking), we haven't looked for keccak's potential vulnerabilities, and we have not gathered usability info from users other than ourselves.

If we agree on this then it would be good if we can add listed items into the "out of the scope" testing section with some rationale.

PR #13, opened 2023-06-20
RFC for branch auto sync tool

dkijania:


0003-branch-auto-sync-tool.md
```diff
@@ -0,0 +1,235 @@
...
+This RFC adresses problem of manual branch merging operations for our major branches: compatible/berkeley/develop
+
+## Motivation
+[motivation]: #motivation
+
+Aim of this project is to satisfy needs for automatic branch synchronization between important branches. So far this is manual process of detecting changes between branches, Pull requests creation, checking if there is no merge conflicts and pushing changes to target branch. This tool automates above process.
+
+## Detailed design
+
+For MVP we want only merge conflicts detection.
```
mrmr1993, line 13: I don't think that this proposal achieves the original goals of the project unfortunately. I'll cross-post my slack comment here for completeness:
> ## Problem statement
> We often have PRs where a change to a branch 'lower down' in the branch stack (`master -> compatible -> rampup -> berkeley -> develop -> (izmir)`) conflict with changes made further up the chain, because code was deleted, some other change was made to the same location, etc. Previously we had to manually 'merge back' these branches and painfully fix the merge conflicts originating from all PRs in one large batch. To reduce the burden of doing this, we added a blocking job to CI to prevent these conflicts from being created, moving the responsibility from myself and others onto the PR's author. This increased friction somewhat for individual PRs, but it has massively reduced the time that is required to keep these branches up-to-date with eachother.
> ## Current state
> As I understand it, the current design of the system operates on the destination branches, automatically opening a PR when the merge conflicts are created in those branches. Currently this isn't very useful due to the blocking CI jobs, since those conflicts can't land and trigger the automation without a force-merge bypass of CI. We could disable these CI jobs, but I'm concerned that this will lead to us accumulating a large number of conflicts again, and that we lose the fidelity of 'your PR conflicts, so you should fix it'.
> ## Original proposal
> The original design that we discussed had 2 components:
> * when a PR lands in an earlier branch in the stack, that change is automatically propagated to all of the later branches via a sequence of merges
>   - I can't remember if we came to a consensus on whether CI should run at each stage, given how long it takes and the inherent raciness with other merges. My instinct says 'no', but I'm probably not the right person to make that decision.
> * when a PR has conflicts with a later branch in the stack, the PR is blocked from merging, and a new PR is created automatically for each(/the next) conflicting branch, where the merge conflicts can be fixed, and with a standardised naming convention. Ideally this would be accompanied by a message on the original PR, so that it is clear to all contributors what the necessary next steps are. As currently, the earlier PRs will block until the later ones are successfully approved and merged.
> 
> The goal of this is to streamline the workflow across the branches, while still ensuring that PR authors are responsible for landing their changes to all the necessary places.

dkijania: @mrmr1993 I received requirements for this tool in a verbal form so probably that was a misunderstanding/miscommunication :( . Tool partially resolves requirements from your message:

This is implemented:

```
when a PR lands in an earlier branch in the stack, that change is automatically propagated to all of the later branches via a sequence of merges
```

While this :

```
when a PR has conflicts with a later branch in the stack, the PR is blocked from merging, and a new PR is created automatically for each(/the next) conflicting branch, where the merge conflicts can be fixed, and with a standardised naming convention. Ideally this would be accompanied by a message on the original PR, so that it is clear to all contributors what the necessary next steps are. As currently, the earlier PRs will block until the later ones are successfully approved and merged.
```

This is still TBD. There is no problem to adding it to the scope. The question i have is do we want to still keep compatibility checks in  build kite for each PR or we want to remove them while automatically creates mentioned PR which are supposed to resolve conflicts?

My humble proposal is to modify only to inform author that a PR was automatically created and provide link to newly created PR by auto sync tool.

mrmr1993: > do we want to still keep compatibility checks in build kite for each PR

Yes. The goal of this is to have PR authors fix their own conflicts. As soon as we let the conflicts creep into the protected branches, we've seen that the conflicts accumulate and somebody has to periodically suffer through the process of reverse-engineering how all of the conflicts should be resolved.

I suppose you could describe what we're after as 'exactly what we have now, with automation'.

mrmr1993 (changes requested):
Marking 'requested changes' to stop GitHub reminders. See previous comments for requested changes.

PR #17, opened 2023-07-06
Foreign field multiplication gate

mitschabaude:
RFC for a revised ffmul gate


0006-ffmul-revised.md
```diff
@@ -0,0 +1,518 @@
...
+Another constraint on the gate layout is that we can only do 4 lookups per row, and we need 7 lookups on the carry chunks $c_{1,0},\ldots,c_{1,72}$. So these have to be divided up between the two rows.
+
+The remaining witnesses are just put into any free cells to define the final gate layout:
+
+| FFmul  | 0p | 1p | 2p | 3p | 4p | 5p | 6p | 7  | 8  | 9  | 10 | 11 | 12 | 13 | 14 |
+| -------| -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- |
+| _Curr_ | $a_0$ | $a_1$ | $a_2$ | $b_0$ | $b_1$ | $b_2$ | $p_{10}$ | $c_{1,0}$ | $c_{1,12}$ | $c_{1,24}$ | $c_{1,36}$ | $c_{1,84}$ | $c_{1,86}$ | $c_{1,88}$ | $c_{1,90}$ | 
+| _Next_ | $r_{01}$ | $r_2$ | $q_0$ | $q_1$ | $q_2$ | $q'_2$ | $p_{110}$ | $c_{1,48}$ | $c_{1,60}$ | $c_{1,72}$ | $c_0$ |  |  |  | 
+
+The limbs of $f' = 2^{3\ell} - f$, which feature in a few of the constraints, are embedded into the gate as coefficients.
```
mrmr1993, line 413: Some constraints (e.g. `C1`) refer to `f` rather than `f'`. Can you update the constraints to use `f'` if we won't be storing `f`, or otherwise propose storing the limbs of `f`?

0006-ffmul-revised.md
```diff
@@ -0,0 +1,518 @@
...
+
+### FFmul gadget
+
+As the soundness proof showed, the ffmul gate is not complete without a number of external checks. The job of an ffmul _gadget_ (a provable method which wraps the gate) is to take care of these checks and provide an API without pitfalls.
+
+As usual, we want the gadget to add all necessary checks to its outputs $q$ and $r$ but not to its inputs $a$ and $b$ (because, if all gadgets fully constrained their inputs, then we would add these constraints multiple times if we used the same inputs in multiple gadgets, which can be very inefficient).
+
+In addition, there should be an advanced flag to skip range checks on $r$, for use cases where $r$ is later asserted to equal a value which is already known to be in the range (for example, the constant 1 when constraining a foreign field inverse).
+
+By default, we will do even more checks on $r$ than required in our soundness proof: We also add a bounds check for $r$ which shows that $r < 2^{2\ell}(f_2 + 1)$. This enables us to use $r$ as input to another ffmul gate or other foreign field operations without extra checks, which is what we expect from a safe API.
```
mrmr1993, line 423: NB: it's safe to use the result directly with the foreign field add gate. We only need to do the bounds check with multiplication because of the extra bound required by CRT

0006-ffmul-revised.md
```diff
@@ -0,0 +1,518 @@
...
+  externalChecks: ExternalChecks,
+  foreignModulus: bigint,
+  skipRemainderCheck?: boolean
+): {
+  q: ForeignField;
+  r: Option<ForeignField>;
+  rCompact: [Field, Field];
+} {
+  // compute witnesses
+  let [q, r01, r2, p10, p11, c0, c1, qBound] = exists(FFMulWitness, () => {
```
mrmr1993, line 440: Nit: use a record here as the return type

0006-ffmul-revised.md
mitschabaude: Good point!

0006-ffmul-revised.md
```diff
@@ -0,0 +1,518 @@
...
+$$\tag{RC: bound for $q$}
+q'_2 \in [0,2^\ell)
+$$
+
+This implies that $a,b,q \in[0, 2^{2\ell}(f_2 + 1))$. For $r$, we already know that $r \in [0, 2^{3\ell})$.
+
+We now get the following upper bound and lower bound:
+
+$$
+ab - qf - r \le ab < 2^{4\ell}(f_2 + 1)^2
```
querolita, line 361: Perhaps let's add a middle `ab ≤ (2^{2\ell} * f_2 + 2^\ell * 2^\ell + 2^\ell )^2`, then that gives you a more accurate upper bound that `ab < 2^{4\ell}(f_2+1)^2 + 2^2\ell + 2*2^\ell*2^\ell*(f_2 + 1)`

0006-ffmul-revised.md
```diff
@@ -0,0 +1,518 @@
...
+In the case that $f < 2^{259}$, we have $(f_2 + 1) \le 2^{259 - 2\ell} = 2^{83}$, and our estimate works out to be 
+
+$$
+|ab - qf - r| < 2^{3\ell} \cdot (2^{254} + 1)
+$$
+
+For $n > 2^{254}$, wich is true for both Pasta moduli, we conclude that
+
+$$
+|ab - qf - r| < 2^{3\ell}n
```
querolita, line 383: In any case, it still holds for the more accurate value.

0006-ffmul-revised.md
```diff
@@ -0,0 +1,518 @@
...
+| _Next_ | $r_{01}$ | $r_2$ | $q_0$ | $q_1$ | $q_2$ | $q'_2$ | $p_{110}$ |  |  |  |  |  |  |  | 
+
+Another constraint on the gate layout is that we can only do 4 lookups per row, and we need 7 lookups on the carry chunks $c_{1,0},\ldots,c_{1,72}$. So these have to be divided up between the two rows.
+
+The remaining witnesses are just put into any free cells to define the final gate layout:
+
+| FFmul  | 0p | 1p | 2p | 3p | 4p | 5p | 6p | 7  | 8  | 9  | 10 | 11 | 12 | 13 | 14 |
+| -------| -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- |
+| _Curr_ | $a_0$ | $a_1$ | $a_2$ | $b_0$ | $b_1$ | $b_2$ | $p_{10}$ | $c_{1,0}$ | $c_{1,12}$ | $c_{1,24}$ | $c_{1,36}$ | $c_{1,84}$ | $c_{1,86}$ | $c_{1,88}$ | $c_{1,90}$ | 
+| _Next_ | $r_{01}$ | $r_2$ | $q_0$ | $q_1$ | $q_2$ | $q'_2$ | $p_{110}$ | $c_{1,48}$ | $c_{1,60}$ | $c_{1,72}$ | $c_0$ |  |  |  |
```
querolita, line 411: Missing `product1_hi_1`

querolita: I placed it in (Next, 7). Here's the layout in the prototype:
```
//~ | col | `ForeignFieldMul`            | `Zero`                  |
//~ | --- | ---------------------------- | ----------------------- |
//~ |   0 | `left_input0`         (copy) | `remainder01`     (copy) |
//~ |   1 | `left_input1`         (copy) | `remainder2`      (copy) |
//~ |   2 | `left_input2`         (copy) | `quotient0`       (copy) |
//~ |   3 | `right_input0`        (copy) | `quotient1`       (copy) |
//~ |   4 | `right_input1`        (copy) | `quotient2`       (copy) |
//~ |   5 | `right_input2`        (copy) | `quotient_bound`  (copy) |
//~ |   6 | `product1_lo`         (copy) | `product1_hi_0`   (copy) |
//~ |   7 | `carry1_0_11`      (plookup) | `product1_hi_1`          |
//~ |   8 | `carry1_12_23`     (plookup) | `carry1_48_59` (plookup) |
//~ |   9 | `carry1_24_35`     (plookup) | `carry1_60_71` (plookup) |
//~ |  10 | `carry1_36_47`     (plookup) | `carry1_72_83` (plookup) |
//~ |  11 | `carry1_84_85`               | `carry0`                 |
//~ |  12 | `carry1_86_87`               |                          |
//~ |  13 | `carry1_88_89`               |                          |
//~ |  14 | `carry1_90`                  |                          |
```

0006-ffmul-revised.md
```diff
@@ -0,0 +1,518 @@
...
+  // (because these need to be in batches of 3 for efficiency)
+  externalChecks.addRangeCheck(qBound);
+
+  // if we're skipping the r checks, only return q and r in compact form
+  if (skipRemainderCheck) {
+    return { q, r: None, rCompact: [r01, r2] };
+  }
+
+  // split up the compact r01 limb to get witnesses for r range check
+  let [r0, r1] = exists([Field, Field], () => splitCompactLimb(r01));
```
querolita, line 460: I guess this is at the Snarky level though

0006-ffmul-revised.md
```diff
@@ -0,0 +1,518 @@
...
+  let r = ForeignField.fromLimbs([r0, r1, r2]);
+
+  // add range check on r (compact mode)
+  addCompactRangeCheck(r, r01);
```
querolita, line 464: I guess this should have `r01` and `r2` as inputs perhaps?

querolita (approved):
I left a few comments to be more accurate in some parts, and add some clarifications somewhere else. But I don't think those comments are sufficient to make the RFC not pass. 

Perhaps the detailed version I sometimes ask for, should be part of the mina book instead, but not the RFC which should be more direct and right to the point. So that's why I am approving after all.

It is great to see someone else's view about this problem. Your collaboration has been key in this process. Thanks for your time.

0006-ffmul-revised.md
mitschabaude: but $ab < 2^{4\ell}(f_2 + 1)^2$ is correct, because $a, b < 2^{2\ell}(f_2 + 1)$ as I showed above. your bound is less tight

mitschabaude: note that $2^\ell a_1 + a_0 \le 2^\ell (2^\ell - 1) + (2^\ell - 1) = 2^{2\ell} - 1$, so estimating that part with $< 2^{2\ell}$ is perfect

0006-ffmul-revised.md
mitschabaude: Needs to have the individual limbs as inputs IMO because it needs those as witnesses

mitschabaude: Except if that function creates the witnesses of course, but I wanted to be as explicit as possible here

0006-ffmul-revised.md
mitschabaude: yes, I was thinking of an OCaml or TS or maybe snarky-rs gadget, similar to this one: https://github.com/MinaProtocol/mina/blob/5121fed0fe0e96a13c9ceca67af1dd763bc4fe34/src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml#L1197

0006-ffmul-revised.md
```diff
@@ -0,0 +1,518 @@
...
+
+$$\tag{$2''''$}
+a b - q f - r = 2^{3\ell} n \delta
+$$
+
+#### Final bound
+
+There is one step remaining to make our gate sound: We must prove that $a b - q f - r$ cannot overflow $2^{3\ell} n$, so $\delta$ is in fact zero.
+
+To do this, our strategy is to add an additional bound on the highest limbs of $a$, $b$, $q$. Skipping the lower limbs and $r$ is fine because they all contribute a negligible amount to the overall estimate.
```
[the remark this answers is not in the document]
mitschabaude: Sure I can add that, in fact I think $2^{259} - 1$ is the precise maximum because you can see in one of the inequalities that the modulus needs to satisfy $2^{2\ell} (f_2 + 1) \le n$, and $2^{2\ell} \lfloor \frac{n}{2^{2\ell}} \rfloor = 2^{259}$

0006-ffmul-revised.md
```diff
@@ -0,0 +1,518 @@
...
+
+Equation $\text{(C1: native)}$ implies that there is an $\varepsilon\in\mathbb{Z}$ such that
+
+$$
+ab - qf - r =  \varepsilon n.
+$$
+
+If the foreign field was small enough, we could prove $|ab - qf - r| < n$ (after adding a few range checks), and we would be finished at this point, because we could conclude that $\varepsilon = 0$. However, for the moduli $f$ we care about, $ab \approx qf \approx f^2$ is much larger than $n$, so we need to do more work.
+
+The broad strategy is to also constrain $(1)$ modulo $2^{3\ell}$, which implies that it holds modulo the product $2^{3\ell}n$ (this the "chinese remainder theorem"). The modulus $2^{3\ell}n$ is large enough that it can replace $n$ in the last paragraph and the argument actually works.
```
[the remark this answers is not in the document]
mitschabaude: I fixed the formulation. Since the proof in the end doesn't invoke the CRT (but proves the part of it we need inline, which I think makes it easier to read -- less magic), I don't think we need any more details on the CRT

0006-ffmul-revised.md
```diff
@@ -0,0 +1,518 @@
...
+
+By design, $c_1 \in [0, 2^{\ell + 3})$. In combination with our estimate on $p_2$ and RCs on $r_2$ and $p_{11} = p_{110} + 2^\ell p_{111}$, we can prove that $\text{(C6: high limb)}$ does not overflow $n$ and holds over the integers:
+
+$$
+|p_2 - r_2 + p_{11} + c_0 - 2^{\ell} c_1| < 6\cdot 2^{2\ell} + 2^{\ell} + 2^{\ell + 2} + 4 + 2^{2\ell + 3} < 2^{2\ell + 4} < n
+$$
+
+Therefore, we can plug in $\text{(C6)}$ into $(2'')$ to prove that
+
+$$\tag{$2'''$}
```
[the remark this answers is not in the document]
mitschabaude: ok used 3, 4 etc!

0006-ffmul-revised.md
mitschabaude: Yeah, maybe we should make the bounds check an independent option

0006-ffmul-revised.md
```diff
@@ -0,0 +1,2249 @@
...
+* For Curve25519 we have $f'_0 = 19$
+* For seqp256k1 we have $f'_0 = 2^{32} + 2^9 + 2^8 + 2^7 + 2^6 + 2^4 + 1$
+
+In these cases, the negative $-|q|_2f'_0$ is small and should never cause underflow for "random" input values. In other words, the prover can use an incorrect output $r$ in almost every foreign field multiplication.
+
+**Conclusion**: The full range check $q_2 \in [0,2^\ell)$ is necessary.
+
+> Counter-example 2: Compact range-check on $q_0 + 2^\ell q_1$
+
+TBC
```
joseandro, line 2248: Our goal is to have a FFmul RFC that will lock its design. For that reason, shouldn't we confirm the TBC before we "lock it" and move on to implementation?

0006-ffmul-revised.md
```diff
@@ -509,10 +520,1725 @@ In summary, we chose the gate design presented in this RFC because:
...
+    - `multi-range-check` $b$
+- Compute and constrain high bounds of inputs $a'_2$ and $b'_2$
+    - `Generic` $a_2 + (2^\ell - f_2 -1) = a'_2$ and $b_2 + (2^\ell - f_2 -1) = b'_2$
+- If $r$ needs to be canonical ($<f$ compute a full bound check: 
+    - `ForeignFieldAdd` $r + 2^t = 1 \cdot f + r'$
+    - `multi-range-check` $r'$
+
+Copy constraints must connect the above witness cells to their respective input cells within the corresponding external check gates witnesses.
+
+## Appendix L: Soundness
```
mitschabaude, line 1839: In my opinion, we don't need this section, because the same proof in a more suitable form is already contained in the main section. I think we should only keep the last part, "necessity of range checks"

0006-ffmul-revised.md
mitschabaude: Yeah I think we can remove the "counter-example 2" section, I feel it's not necessary to put in the work of constructing another counter-example

0006-ffmul-revised.md
querolita: Yeah, probably I will just link it inside "Prior art", and just extract the necessity of range checks section, which is mainly the reason why we are switching the design.

0006-ffmul-revised.md
joseandro: Sounds good, thank you all!

0006-ffmul-revised.md
```diff
@@ -0,0 +1,1932 @@
...
+
+### Naïve approach
+
+Naïvely, this implies that we have elements of size $f^2 - 1$ that must split them up into limbs of size at most $n - 1$.  For example, if the foreign field modulus is $256$ and the native field modulus is $255$ bits, then we'd need $\log_2((2^{256})^2 - 1) \approx 512$ bits and, thus, require $512/255 \approx 3$ native limbs.  However, each limb cannot consume all $255$ bits of the native field element because we need space to perform arithmetic on the limbs themselves while constraining the foreign field multiplication.  Therefore, we need to choose a limb size that leaves space for performing these computations.
+
+Later in this document (see the section entitled "Choosing the limb configuration") we determine the optimal number of limbs that reduces the number of rows and gates required to constrain foreign field multiplication.  This results in $\ell = 88$ bits as our optimal limb size.  In the section about intermediate products we place some upperbounds on the number of bits required when constraining foreign field multiplication with limbs of size $\ell$ thereby proving that the computations can fit within the native field size.
```
mrmr1993, line 563: The section is just called `limb configuration` now

0006-ffmul-revised.md
```diff
@@ -0,0 +1,1932 @@
...
+
+As stated above, we want the foreign field modulus to fit in as few rows as possible and we need to constrain operands $a, b$, the quotient $q$ and remainder $r$. Each of these will require cells for each limb.  Thus, the number of cells required for these is
+
+$$
+cells = 4 \cdot limbs
+$$
+
+It is highly advantageous for performance to constrain foreign field multiplication with the minimal number of gates. This not only helps limit the number of rows, but also to keep the gate selector polynomial small. Because of all of this, we aim to constrain foreign field multiplication with a single gate (spanning at most $2$ rows). As mentioned above, we have a maximum of 14 permuteable cells per gate, so we can compute the maximum number of limbs that fit within a single gate like this.
```
mrmr1993, line 723: > keep the gate selector polynomial small
What does this mean? Sentence should probably be deleted

mrmr1993: Suggestion: replace the whole final sentence with
> We can compute the maximum number of limbs that fit in the 14 permutable columns as follows:

0006-ffmul-revised.md
```diff
@@ -0,0 +1,1932 @@
...
+$$
+\begin{aligned}
+x - y  &  = 0 \\
+x &= y,
+\end{aligned}
+$$
+
+which is a contradiction with $x < y$.
+
+## Appendix D: Gate constraints
```
mrmr1993, line 802: This whole section feels redundant with the main body. Would be better to replace this whole section with just the diagrams, and refer to them from the relevant points.

0006-ffmul-revised.md
```diff
@@ -0,0 +1,1932 @@
...
+Note that $r'_2$ must be range checked by a `multi-range-check` gadget. Then, the exteral checks structure will keep track of it and make "dense" multi range check with triples of remainders coming from a number of `ForeignFieldMul` from a larger circuit. 
+
+- Store a copy of the limbs $r_{01}$ and $r_2$ in its witness
+- Compact-range check that each limb is $\ell$ bits each
+- Decompose it into $r_0, r_1, r_2$ in the compact range check gadget
+- Compute $r'_2$ using a `Generic` gate
+- Push the bound to a list of external checks to be multi range checked together.
+
+
+### Inlining compact full bound check inside a gate
```
mrmr1993, line 1319: This feels like it would be better suited as a separate, dedicated document

0006-ffmul-revised.md
```diff
@@ -0,0 +1,2249 @@
...
+
+We will now derive the constraints necessary to prove that
+
+$$\tag{1}
+a b = qf + r
+$$
+
+At the same time, we will prove soundness -- i.e., the constraints we add imply $(1)$. 
+
+> In the following, we will use the numbering used in the implementation, which might not correspond to the chronological order where they appear explained here in the text.
```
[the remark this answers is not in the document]
querolita: Sure, the code was using a different ordering than the "natural" sense to be able to trigger some individual constraints that are less restrictive than others. But sure, let's keep it simple and "chronological" order in the RFC for readability.

0006-ffmul-revised.md
```diff
@@ -0,0 +1,2249 @@
...
+
+Let $n$ be the _native_ field modulus (for example, $n$ is one of the Pasta primes). Our first constraint is to check $(1)$ modulo $n$:
+
+$$\tag{C5: native}
```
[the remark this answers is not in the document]
querolita: It is referring to the prototype order, where I have been testing some constraints still hold / still are triggered.

0006-ffmul-revised.md
```diff
@@ -0,0 +1,2249 @@
...
+ab = qf + r\mod n
```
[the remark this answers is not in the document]
querolita: oh that was just to make it clearer to read, but then a few lines later it is explained that the negated modulus is used. In any cae, I clarified that sooner in text.

0006-ffmul-revised.md
```diff
@@ -0,0 +1,2249 @@
...
+
+To make them available, they have to go in permutable cells. Kimchi has 7 permutable cells per row, so we exactly fit these values in the current and next row of a single-row gate.
+
+
+| FFmul  | 0p | 1p | 2p | 3p | 4p | 5p | 6p |
+| -------| -- | -- | -- | -- | -- | -- | -- |
+| _Curr_ | $a_0$ | $a_1$ | $a_2$ | $b_0$ | $b_1$ | $b_2$ | $p_{10}$ |
+| _Next_ | $r_{01}$ | $r_2$ | $q_0$ | $q_1$ | $q_2$ | $q'_2$ | $p_{110}$ |  |  |  |  |  |  |  | 
+
+Another constraint on the gate layout is that we can only do 4 lookups per row, and we need 7 lookups on the carry chunks $c_{1,0},\ldots,c_{1,72}$. So these have to be divided up between the two rows. In order to reuse the same `LookupPattern` in both rows of the gadget, we must place a value that is known to be less than 12 bits in the same column of the `Next` row (or leave it empty with a zero).
```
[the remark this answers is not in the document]
querolita: Oh that's a great idea, I will add this now.

0006-ffmul-revised.md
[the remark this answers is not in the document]
mitschabaude: Yeah actually the implementation should use $f = 2^{3\ell} - f'$, I spelled that out in the gate layout section. It just seemed inconvenient to the text flow to introduce $f'$ at this point 😅

0006-ffmul-revised.md
```diff
@@ -0,0 +1,2249 @@
...
+  externalChecks.addRangeCheck(rBound);
+
+  // return q, r, rCompact
+  return { q, r: Some(r), rCompact: [r01, r2] };
+}
+```
+
+### Spec
+
+[WIP Kimchi PR with spec change](https://github.com/o1-labs/proof-systems/blob/ffmul/rangecheck/book/src/specs/kimchi.md#foreign-field-multiplication)
```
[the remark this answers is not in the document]
mitschabaude: The idea (mentioned in the RFC template under "Ephemeral details") was to link to a PR with spec changes from within the RFC. But since that code spec doesn't include more info than this RFC, I'm fine to delete

0006-ffmul-revised.md
querolita: > keep the gate selector polynomial small
I guess this refers to having  a gate that has small degree, but yeah not super clear. Delete.

0006-ffmul-revised.md
querolita: Yeah, that's true. I will then keep diagrams only.

0006-ffmul-revised.md
querolita: Deleting for now

0006-ffmul-revised.md
querolita: I found some diagrams and equations to go well with a bit of text. I reduced the amount of sections to only keep the ones containing actual diagrams and "novel" content.

PR #18, opened 2023-07-06
gpu compute

fabrizio-m:


xxxx-gpu-msm.md
```diff
@@ -0,0 +1,92 @@
...
+GPU based MSM implementation.
+
+## Summary
+
+A rust library implementing gpu based acceleration for computing MSM, it can be integrated in kimchi and
+other projects that make use of MSMs with the pasta curves.
+
+## Motivation
+
+One of the mains components of kimchi's runtime is the multi scalar multiplication (MSM), it consist of
```
mrmr1993, line 12: Nit: `s/mains/main`

xxxx-gpu-msm.md
```diff
@@ -0,0 +1,92 @@
...
+This is a crate implementing general field arithmetic, general curve arithmetic and specifically the
+pasta curves and their field. It uses no dependencies and avoids hardware features that are unsupported
+by GPUs in general and WGSL in particular.
+The code is also written in a way optimized for GPUs, but it can still be used in any architecture, while
+it will be generally slower that other cpu implementations it can be used for things like tests.
+
+### shader
+
+The shader is a small crate that will be compiled to WGSL, the code is very simple, mostly calls to
+functions from the arithmetic crate, compilation takes a significative time and thus the compiled
```
mrmr1993, line 40: Nit: `s/significative/significant`

xxxx-gpu-msm.md
```diff
@@ -0,0 +1,92 @@
...
+shader will be saved and the runner will used that precompiled shader by default.
+
+### runner
+
+This is library that will be used by end users of the crate, it is the part that will run in the CPU
```
mrmr1993, line 45: Nit: I think `end-users` is hypenated

xxxx-gpu-msm.md
```diff
@@ -0,0 +1,92 @@
...
+and will talk to the GPU, it will expose a simple API that will ask for the points that make the base
```
mrmr1993, line 46: Can you add a tiny sketch of the API inline?

xxxx-gpu-msm.md
```diff
@@ -0,0 +1,92 @@
...
+of the MSM and then perform the initialization and provide some struct that can be used to send scalar
+and get back the MSM of them and the base sent in the initialization.
+
+## Test plan and functional requirements
```
mrmr1993, line 50: Can you add a note here to add a comparative benchmark? That seems like the main deliverable

xxxx-gpu-msm.md
```diff
@@ -0,0 +1,92 @@
...
+One of the mains components of kimchi's runtime is the multi scalar multiplication (MSM), it consist of
+basically taking a list of points, adding each point to itself N times, and then add all the resulting
+points together.
+This library will accelerate that process doing the computation in a gpu, it will also serve as a reference
+point to measure the improvements that gpu computations can provide us.
+
+This will build on top of an already existing implementation, and thus some of the design choices will be
+based on the fact that the particular decision is cheap to implement because it is already implemented.
+
+## Detailed design
```
mrmr1993, line 21: This section doesn't seem to mention the integration with kimchi in any detail. Can you add something about how that will look, what we require in terms of feature flags / feature detection /  etc.?

xxxx-gpu-msm.md
```diff
@@ -0,0 +1,92 @@
...
+### runner
+
+This is library that will be used by end users of the crate, it is the part that will run in the CPU
+and will talk to the GPU, it will expose a simple API that will ask for the points that make the base
+of the MSM and then perform the initialization and provide some struct that can be used to send scalar
+and get back the MSM of them and the base sent in the initialization.
+
+## Test plan and functional requirements
+
+Given that the goal is more to see the kind of performance gains we can obtain, the testing can be
```
mrmr1993, line 52: I don't think we can test it easily in CI, right? We probably don't need a solution now, but worth calling out that it'll be manual testing, that we need to keep explicitly running it to make sure it doesn't break, etc.

xxxx-gpu-msm.md
```diff
@@ -0,0 +1,92 @@
...
+amount of reviewing and additional test considering that the arithmetic crate is an implementation
+from scratch of the basic cryptographic building blocks.
+There are already some unit tests and some tests using formal verification, but the formal verification
+tests take a while to run a probably can't cover all cases.
+
+## Drawbacks
+
+- There is the possibility of something outside of our control not working, like the nvidia Vulkan drivers
+  when I tried to use Vulkan instead of wgpu.
+- There could be no significative gains, the Vulkan implementations was able to run ~3 times faster than
```
mrmr1993, line 64: Nit: `significant`

xxxx-gpu-msm.md
```diff
@@ -0,0 +1,92 @@
...
+
+Given that the goal is more to see the kind of performance gains we can obtain, the testing can be
+limited to asserting that the result obtained is the same as our current CPU implementation.
+Before this becomes the default implementation to be used in production we will require a considerable
+amount of reviewing and additional test considering that the arithmetic crate is an implementation
+from scratch of the basic cryptographic building blocks.
+There are already some unit tests and some tests using formal verification, but the formal verification
+tests take a while to run a probably can't cover all cases.
+
+## Drawbacks
```
mrmr1993, line 60: The details in this section feel like they belong more in the 'functional requirements' above. E.g. `compatibility with nvidia Vulkan is out-of-scope for this prototype, due to driver issues`

xxxx-gpu-msm.md
fabrizio-m: Can we maybe test it in CI with integrated graphics or a software renderer? 
As implemented right now it will take the best available compatible device, if there is something like `llvmpipe` it should be used automatically.

xxxx-gpu-msm.md
fabrizio-m: added

xxxx-gpu-msm.md
fabrizio-m: added

xxxx-gpu-msm.md
fabrizio-m: added

xxxx-gpu-msm.md
fabrizio-m: It was just an example, the reason Vulkan is not a target is just because it won't work in browser, we can maybe have a non-browser specialized version that uses Vulkan after this prototype.
Do you still want some change to either section?

xxxx-gpu-msm.md
mrmr1993: I worry that if we fallback, we may end up not explicitly testing it. I don't think e.g. GitHub actions have GPU support.

xxxx-gpu-msm.md
mrmr1993: It's just a nit really, it's probably fine to leave as-is

mitschabaude (approved):
Looks like a good plan to me!

mrmr1993 (approved):

mrmr1993 (commented):
@fabrizio-m this seems ready to merge; can you add an RFC number?

PR #22, opened 2023-07-13
Small fix to ffmul RFC

mitschabaude:
This fixes an error in stating the first constraint which slipped in during addressing reviews

### DIFF
--- a/0006-ffmul-revised.md
+++ b/0006-ffmul-revised.md
@@ -1296,7 +1296,7 @@ Compute and constrain the native modulus values, which are used to check the con
 - $r_n = 2^{2\ell} \cdot r_2 + r_{01} \mod n$
 - $f_n = 2^{2\ell} \cdot f_2 + 2^{\ell} \cdot f_1 + f_0 \mod n$
 
-Actually, there is no need to store in the coefficients the limbs of $f$, but only $(f'_0, f'_1, f'_2)$ and $f_2$. Then the native constrain uses the negated limbs of `f` instead.
+Actually, there is no need to store in the coefficients the limbs of $f$, but only $(f'_0, f'_1, f'_2)$ and $f_2$. Then the native constrain uses the negated limbs of `f` instead and subtracts the additional $2^{3\ell}q_n$ term.
 
 ### Decompose middle intermediate product
 
@@ -1352,7 +1352,7 @@ And some more range checks. In order to save one space in permutable cells, and 
 Using the checked native modulus computations we constrain that
 
 $$
-a_n \cdot b_n + q_n \cdot f'_n - r_n = 0 \mod n.
+a_n \cdot b_n + q_n \cdot f'_n - r_n - 2^{3\ell}\cdot q_n = 0 \mod n.
 $$
 
 ### Decompose the higher quotient bound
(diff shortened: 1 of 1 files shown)


0006-ffmul-revised.md
```diff
@@ -58,14 +58,8 @@ Note that the actual order of the constraints in the code might not correspond t
...
 
-$$
-ab = qf + r\mod n
-$$
-
-or equivalently, which only uses the negated limbs of the foreign modulus:
-
 $$\tag{C1: native}
-ab + q f' - r = 0 \mod n
+ab = qf + r\mod n
```
mitschabaude, line 62: replacing $ab = qf + r$ with $ab + qf' - r = 0$ here was wrong because:
1. it was missing a term $2^{3\ell}q$
2. it used $f'$ which wasn't defined in the text at that point

mitschabaude: We explain later in the text that the gate implementation uses $2^{3\ell} - f'$ in place of $f$. This is sufficient.

querolita: I added some other fixes in other parts of the text, I realized they were wrong while doing the implementation.

mrmr1993: Can you move the definition of `f'`, and inline the `f'` version? The constraints that we enumerate here should be exactly what we're intending to have in the code, there should be no interpretation necessary to check the RFCs constraints against the code.

0006-ffmul-revised.md
```diff
@@ -1302,7 +1296,7 @@ Compute and constrain the native modulus values, which are used to check the con
 - $r_n = 2^{2\ell} \cdot r_2 + r_{01} \mod n$
 - $f_n = 2^{2\ell} \cdot f_2 + 2^{\ell} \cdot f_1 + f_0 \mod n$
 
-Actually, there is no need to store in the coefficients the limbs of $f$, but only $(f'_0, f'_1, f'_2)$ and $f_2$. Then the native constrain uses the negated limbs of `f` instead.
+Actually, there is no need to store in the coefficients the limbs of $f$, but only $(f'_0, f'_1, f'_2)$ and $f_2$. Then the native constrain uses the negated limbs of `f` instead and subtracts the additional $2^{3\ell}q_n$ term.
```
mrmr1993, line 1299: This crucial detail of how to implement the native constraint is nearly at the end of the appendix. This is not good.

mitschabaude: Actually we also spell out this detail in the gate layout section (main part of the RFC):

> The limbs of $f' = 2^{3\ell} - f$, which feature in a few of the constraints, are embedded into the gate as coefficients. The $\text{(C1)}$ constraint features $f$, but the gate implementation expands it out as $f = 2^{3\ell} - f' = 2^{3\ell} - (f_0' + 2^\ell f_1' + 2^{2\ell} f_2')$ to be able to use the same coefficients

0006-ffmul-revised.md
mitschabaude: ok that makes sense!

mrmr1993 (approved):

PR #27, opened 2023-08-28
Optimized Keccak for zkVM 

querolita:
This RFC presents a design of a Kimchi chip for the Keccak hash function. It is optimized for the zkVM project, making use of our [generalized expression framework](https://file+.vscode-resource.vscode-cdn.net/Users/querolite/WORKSPACE/rfcs/0007-generalized-expression-framework.md), and assuming a variant of Kimchi using a Nova-styled folding over BN128 and KZG commitments.

### DIFF
--- a/0008-zkvm-keccak.md
+++ b/0008-zkvm-keccak.md
@@ -1,0 +1,569 @@
+# zkVM-Optimized Keccak
+
+This RFC presents a design of a Kimchi chip for the Keccak hash function. It is optimized for the zkVM project, making use of our [generalized expression framework](./0007-generalized-expression-framework.md), and assuming a variant of Kimchi using a Nova-styled folding over BN128 and KZG commitments.
+
+## Summary
+
+This document presents a design to deeply support the Keccak hash function for efficient pre-image oracle access within the proof system. To do so, it leverages a *bitwise sparse representation* to compute boolean operations (XORs, ANDs), and sequences of them.
+
+## Motivation
+
+One component of the Optimism stack system is the pre-image data oracle which is responsible for getting information about the EVM into the MIPS world. The OP stack currently uses the existing ledger which is built on top of Keccak hashes as a baseline.
+
+Having a dedicated chip inside Kimchi to prove correct computation of Keccak hashes from pre-image data (blocks) will result in a more efficient design of the MIPS zkVM as a whole, meeting the performance goals of the [OP RFP](https://github.com/ethereum-optimism/ecosystem-contributions/issues/61#issuecomment-1611488039).
+
+As indicated in the [Keccak PRD](https://www.notion.so/minaprotocol/PRD-Optimized-keccak256-hashing-for-MIPS-VM-33c142f38467464a9b7c21184544dd1d) any member of the OP ecosystem who wanted to prove the state transition between two blocks of an OP stack chain executing the Cannon program in the O(1) MIPS zkVM should complete within 2 days (before optimizations) to meet the latency estimates from the RFP. 
+
+The ultimate objective is to improve the performance of our current Keccak gadget introduced during the Ethereum primitives epic, to get over one hash per second. This is one of the [OKRs of H2 2023](https://www.notion.so/minaprotocol/8dfc798e277447ea860e747b026d9698?v=5daef538906d44db9d12e2e0fbbd4665). 
+The new design would be a success if the number of hashes per second is faster than that of the current [SnarkyML Keccak PoC](https://github.com/MinaProtocol/mina/pull/13196). 
+
+The optimizations in the proposed design rely on a more efficient representation of the `Xor64` operation (i.e. a gadget for 64-bit XOR). Targetting this boolean operation makes sense, as this is the most used component inside the permutation function of Keccak. In particular, the main source of overhead in the current design corresponds to this operation.
+
+## Detailed design
+
+The Keccak hash function (later standardized by NIST as SHA3, with small variations) is known to be a quantum-safe, and very efficient hash to compute. Nonethelss, its intensive use of boolean operations which are costly in finite field arithmetic makes it a SNARK-unfriendly function. In this section, we propose an optimized design for our proof system that should outperform our current [Keccak PoC](https://www.notion.so/minaprotocol/Keccak-gadget-PoC-PRD-59b024bce9d5441c8a00a0fcc9b356ae) to become usable in the zkVM for the OP stack.
+
+### Background
+
+This subsection covers a thorough description of the Keccak algorithm and specifies the required parameters needed for compatibility in our usecase. The reader shall skip this section to find the details of the actual proposal in this RFC. 
+
+The configuration of Keccak must be the same as that of the Ethereum EVM, meaning specifically:
+
+> * [pre-NIST variant of Keccak](https://keccak.team/files/Keccak-reference-3.0.pdf) (uses the $10^*1$ padding rule)
+> * variable input **byte** length (not any bit length)
+> * output length of 256 bits
+> * input and output in big endian
+> * state composed of 25 (5x5) 64-bit words
+> * making a block length of 1600 bits ($b$)
+> * 512 bits capacity ($c$)
+> * 1088 length bit rate ($r := b - c$)
+> * 24 rounds of permutation
+
+The high level idea of this sponge-based algorithm iteratively is to apply a permutation function to an initial state for a number of rounds (absorb phase) and to finally crop the state, when needed, obtaining the digest (squeeze) as follows:
+
+<center>
+
+![](https://upload.wikimedia.org/wikipedia/commons/7/70/SpongeConstruction.svg)
+
+</center>
+
+The following represents the description of the Keccak algorithm tailored to our usecase configuration presented above.
+
+1. **Input:** message
+2. **Requirement:** message is a byte string in big endian
+    - If the input was given as bits and its length was not a multiple of 8 bits, represent the MSB as a full byte before the next step.
+3. **Pad:** pad the message with the padding rule $10^*1$
+    - Append the byte `0x01`. 
+    - Afterwards, append as many `0x00` bytes needed until the length of the resulting message is a multiple of the bit rate (1088 bits). 
+    - XOR the last byte with `0x80`. 
+    > **Edge cases:** input messages whose length is
+    > - _0:_ reject, the "smallest" message is `0x00` which takes 8 bits in length, and will need 1080 padding bits.
+    > - _(1088n) a multiple of the bit rate:_ the padding will take a whole new block of another 1088 bits.
+    > - _(1088n - 1) just below the bit rate:_ the padding will take one single byte `0x81`.
+4. **Initialization:** create a root state $s$ of 1600 bits set to zero.
+    - For convenience, store it as a 5x5 matrix with 64-bit words in each cell.
+5. **Notation:** 
+    - `s[i][j][k]` refers to the k-th least significant byte of the i-th row of the j-th column of the state matrix. 
+    - `message[n]` refers to the n-th most significant byte of the bytestring (big endian).
+6. **Absorb:** for each block of 136 bytes (1088 bits) in the padded bytestring
+    - Pad with 64 `0x00` bytes until reaching a total of 1600 bits.
+    - Map the resulting bytestring to state format
+    - Update the state xoring it with the current block state.
+        - Meaning either the root state for the first block, or the previous state for the rest of blocks.
+    - Apply the permutation function to the state.
+7. **Stringify:** map the state to a bytestring and store the first 1088 bits in an output bytestring.
+8. **Squeeze:** while the bitlength of the resulting bytestring is less than the desired output bit
+    - Apply the permutation function to the state
+    - Map the resulting state to a bytestring
+    - Append the first rate number of bits to the output bytestring
+9. **Hash:** Take the first 256 bits (32 bytes) of the output
+10. **Return:** The hash digest obtained above
+
+> **NOTE** The squeeze step does not take place with our output length of 256 bits and 1088 bit rate.
+
+In order to perform the above steps correctly, mapping bytes into the state in the right order is key:
+
+- `from_bytes_to_state`: the absorb step requires a mapper from a bytestring into the state format.
+- `from_state_to_bytes`: the squeeze steps requires a mapper from a state into the bytestring format.
+
+In particular, this is the order that should be followed. On input a list of bytes in big endian format (the first element corresponds to the left-most byte in a string in hexadecimal format), the first element is mapped to `M[0][0][0]` (the least significant byte of the word in the first column and first row). The following 7 bytes are mapped to the subsequent bytes of `M[0][0]`. The next 8 bytes will go to the cell in the first column, second row `M[1][0]`. The following table illustrates the order of the 25 64-bit words in the state: 
+
+<center>
+
+| x \ y | 0 | 1 |  2 |  3 |  4 |
+| ----- | - | - | -- | -- | -- |
+| **0** | 0 | 5 | 10 | 15 | 20 |
+| **1** | 1 | 6 | 11 | 16 | 21 |
+| **2** | 2 | 7 | 12 | 17 | 22 |
+| **3** | 3 | 8 | 13 | 18 | 23 |
+| **4** | 4 | 9 | 14 | 19 | 24 |
+
+</center>
+
+Then, within each word, the first bytes of the bytestring get mapped to the least significant bytes of the word in the state as:
+
+
+
+More formally, the sequence to convert from bytestring to state follows the pseudocode:
+
+```rust
+for y in [0..5) { // for each column
+    for x in [0..5) { // for each row
+        for z in [0..8) { // for each byte in the word
+            index = 8 * (5 * y + x) + z
+            state[x][y] |= (byte[index]) << 8 * z
+        }
+    }
+}
+```
+
+The same strategy (but reverse order) shall be followed to convert from the state back to bytestring at the end of the sponge, meaning:
+
+```rust
+for y in [0..5) { // for each column
+    for x in [0..5) { // for each row
+        for z in [0..8) { // for each byte in the word
+            byte.push(((state[x][y] >> 8 * z) & 0xFF));
+        }
+    }
+}
+```
+
+#### Permutation function
+
+The main responsible for the security guarantees (pre-image and collision resistance) is the permutation function that is executed inside the sponge. It applies the same algorithm iteratively to a state, for 24 rounds.
+
+> This number comes from the formula $12+2·\ell$ where $\ell=6$ in our case with ($2^6 =$)64-bit words. 
+
+```rust
+fn permutation(state) -> state {
+    for r in [0..24) {
+        state <- round(state, r)
+    }
+    return state
+}
+```
+
+The round function itself is the concatenation of a number of simple algorithms whose duty is to randomize the state in a non-reversible manner. The notation choice here is meant to represent the letters chosen in the official docs of Keccak as closely as possible.
+
+```rust
+fn round(state_in, r) -> state_out {
+    A <- state_in
+    E <- theta(A)
+    B <- pi_rho(E)
+    F <- chi(B)
+    D <- iota(F, r)
+    return D
+}
+```
+
+In the following, we will show the pseudocode of each of these algorithms. Note that positions in X and Y coordinates must be computed modulo 5.
+> Make sure that your language performs the remainder operation instead of modulo. Otherwise, make sure to add `+5` in subtractions to avoid negative indices.
+
+```rust
+fn theta(state_a) -> state_e {
+    A <- state_a
+    for x in [0..5) {
+        C[x] <- A[x][0] xor A[x][1] xor A[x][2] xor A[x][3] xor A[x][4]  
+    } 
+    for x in [0..5) {
+        D[x] <- C[x-1] xor rot(C[x+1], 1) // Left rotation by 1 bit (towards more significant positions)
+        for y in [0..5) {
+            E[x][y] <- A[x][y] xor D[x]
+        }
+    }
+    return E
+}
+```
+
+```rust
+fn pi_rho(state_e) -> state_b {
+    E <- state_e
+    for x in [0..5) {
+        for y in [0..5) {
+            B[y][2x+3y] <- rot(E[x][y], OFF[x][y])
+        }
+    }
+    return B
+}
+```
+
+```rust
+fn chi(state_b) -> state_f {
+    B <- state_b
+    for x in [0..5) {
+        for y in [0..5) {
+            F[x][y] <- B[x][y] xor (not B[x+1][y] and B[x+2][y])
+        }
+    }
+    return F
+}
+```
+
+```rust
+fn chi(state_f, r) -> state_g {
+    G <- state_f
+    G[0][0] <- F[0][0] xor RC[r]
+    return G
+}
+```
+
+#### Constants
+
+Tha algorithms above make use of two sets of constants: 
+- rotation table `OFF[x][y]` is a 5x5 matrix defining a constant rotation offset for each word in the state. For 64-bit words, it corresponds to:
+
+<center>
+
+| x \ y |  0 |  1 |  2 |  3 |  4 |
+| ----- | -- | -- | -- | -- | -- |
+| 0     |  0 | 36 |  3 | 41 | 18 |
+| 1     |  1 | 44 | 10 | 45 |  2 |
+| 2     | 62 |  6 | 43 | 15 | 61 |
+| 3     | 28 | 55 | 25 | 21 | 56 |
+| 4     | 27 | 20 | 39 |  8 | 14 |
+
+</center>
+
+- round constants `RC[r]` in our case is a 24-length array of 64-bit values which are defined as the output of the binary linear feedback shift register (LFSR) given by
+
+$$RC[r] = (x^r \mod{x^8 + x^6 + x^5 + x^4 + 1} )\mod{x} $$
+
+<center>
+
+| `r` |                 `RC` |
+| --- | -------------------- |
+|  0  | `0x0000000000000001` |
+|  1  | `0x0000000000008082` |
+|  2  | `0x800000000000808A` |
+|  3  | `0x8000000080008000` |
+|  4  | `0x000000000000808B` |
+|  5  | `0x0000000080000001` |
+|  6  | `0x8000000080008081` |
+|  7  | `0x8000000000008009` |
+|  8  | `0x000000000000008A` |
+|  9  | `0x0000000000000088` |
+| 10  | `0x0000000080008009` |
+| 11  | `0x000000008000000A` |
+| 12  | `0x000000008000808B` |
+| 13  | `0x800000000000008B` |
+| 14  | `0x8000000000008089` |
+| 15  | `0x8000000000008003` |
+| 16  | `0x8000000000008002` |
+| 17  | `0x8000000000000080` |
+| 18  | `0x000000000000800A` |
+| 19  | `0x800000008000000A` |
+| 20  | `0x8000000080008081` |
+| 21  | `0x8000000000008080` |
+| 22  | `0x0000000080000001` |
+| 23  | `0x8000000080008008` |
+
+</center>
+
+### Bitwise-sparse representation
+
+Here, we present the bitwise-sparse representation of bitstrings that will be used to obtain more efficient computations of boolean functions within our proof system. 
+
+The goal is to represent these operations in finite field algebra more efficiently than they were in the [Keccak PoC](https://www.notion.so/minaprotocol/Keccak-gadget-PoC-PRD-59b024bce9d5441c8a00a0fcc9b356a), for which we took a naive approach where small values occupied full field elements; wasting a lot of empty space. For instance, each row could only xor 16 bits at a time, whereas the field could fit 255 bits. 
+
+Given that simple binary operations are very costly within a SNARK, the new idea consists on arithmetizing boolean operations as much as possible, becoming more efficient with finite field algebra. We introduce the _bitwise-sparse representation_ of bitstrings for this goal.
+
+With this notation, input bitstrings are expanded inserting three `0` bits in between "real" bits (hence the _sparsity_). The motivation for this, is to use intermediate bits to store auxiliary information such as carry bits, resulting from algebraic computations of boolean functions. That means, for each bit in the original bitstring, its expansion will contain four bits. The following expression represents such mapping:
+
+$$\forall X \in \{b_i\}^\ell, b\in\{0,1\}: expand(X) \to 0 \ 0 \ 0 \ b_{\ell-1} \ . \ . \ . \ 0 \ 0 \ 0 \ b_0$$
+
+Even though in each field element of ~254 bits long, one could fit up to 63 real bits with this encoding, each witness value will only store 16 such real bits. This is because the way to perform this mapping is through a 16-bit lookup table (of $2^{16}$ entries, larger than that would be too costly). Then, the expansion of the full 64 bit words could be represented by composition of each of the four 16-bit parts. But this will not be computed in the circuit, since the expansion would not fit in the field.
+
+Let $sparse(X)$ refer to a representation of $X$, where only the indices in the $4i$-th positions correspond to real bits, and the intermediate indices can contain auxiliary information. Concretely,
+
+$$\forall X\in\{b_i\}^\ell:\quad X \equiv \sum_{0}^{\ell-1} 2^i \cdot sparse(X)_{4i} $$
+
+Another set of functions will be defined to be used in this new design. For $i\in\{0,3\}$, let $shift_i(X)$ be the function that performs the XOR operation of the sparse representation of the word $2^{\ell}-1$ ("all-ones" word in binary), shifted $i$ bits to the left, with the input word $X$. More formally,
+
+$$\forall X \in \{b_i\}^\ell, i\in\{0,3\}: shift_i(X) := \left(\ expand(2^\ell - 1)<<_i 0\ \right)\ \oplus \ X$$ 
+
+meaning, on input $expand(X)$, all four of:
+
+$$ 
+\begin{align*}
+shift_0\left(expand(X)\right) := 0 \ 0 \ 0 \ 1_{\ell-1} \ .\ .\ .\ 0 \ 0 \ 0 \ 1_{0} \ \oplus\ 0 \ 0 \ 0 \ b_{\ell-1} \ . \ . \ . \ 0 \ 0 \ 0 \ b_0 \\
+shift_1\left(expand(X)\right) := 0 \ 0 \ 1_{\ell-1} \ 0 \ . \ . \ . \ 0 \ 0 \ 1_{0} \ 0 \ \oplus\ 0 \ 0 \ 0 \ b_{\ell-1} \ . \ . \ . \ 0 \ 0 \ 0 \ b_0 \\
+shift_2\left(expand(X)\right) := 0 \ 1_{\ell-1} \ 0 \ 0 \ . \ . \ . \ 0 \ 1_{0} \ 0 \ 0 \ \oplus\ 0 \ 0 \ 0 \ b_{\ell-1} \ . \ . \ . \ 0 \ 0 \ 0 \ b_0 \\
+shift_3\left(expand(X)\right) := \ 1_{\ell-1} \ 0 \ 0 \ 0 \ . \ . \ . \ 1_{0} \ 0 \ 0 \ 0 \oplus\ 0 \ 0 \ 0 \ b_{\ell-1} \ . \ . \ . \ 0 \ 0 \ 0 \ b_0 \\
+\end{align*}
+$$
+
+The effect of each $shift_i$ is to negate the $4i$-th bits of the expanded output, while leaving the rest of bits unchanged. That implies that one can rewrite the input as:
+
+$$X \iff shift_0(X) + 2 \cdot shift_1(X) + 2^2 \cdot shift_2(X) + 2^3\cdot shift_3(X) ???$$
+
+<details>
+<summary>TODO</summary>
+The above is wrong. Ask Matthew to explain it differently. Why are shifts needed in the first place and we don't just use XOR as explained below?
+</details>
+
+> Checking the correct form of the shifts should be done with a single-column lookup table with just the expanded values for the check, since the non-sparse pre-image is non-relevant here.
+
+Given that the new field size is 254 bits, the whole 64-bit word expansion cannot fit in one single witness element. Instead, they will be split into quarters of 64 bits each (corresponding to real 16 bits each), following this notation:
+
+$$\forall i \in \{0, 3\}:\quad split_i(sparse(X)) := sparse(X_{16i}^{16i+15})$$
+
+Meaning that $\forall X\in(0,2^{64})$:
+
+$$ 
+\begin{align*}
+sparse(X) \equiv &split_0(sparse(X)) + 2^{64} \cdot split_1(sparse(X)) + \\
+& 2^{128}\cdot split_2 (sparse(X)) + 2^{192} \cdot split_3(sparse(X))
+\end{align*}
+$$
+
+#### XOR
+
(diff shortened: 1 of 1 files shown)


0008-zkvm-keccak.md
```diff
@@ -0,0 +1,791 @@
...
+
+Let $sparse(X)$ refer to a representation of $X$, where only the indices in the $4i$-th positions correspond to real bits, and the intermediate indices can contain auxiliary information. Connecting to the above, after performing a series of boolean operations to the $expand(X)$ one can obtain some $sparse(X)$, but both encode the same $X$. Concretely,
+
+$$\forall X\in\{b_i\}^\ell:\quad X \equiv \sum_{0}^{\ell-1} 2^i \cdot sparse(X)_{4i} $$
+
+Given that the new field size is 254 bits, the whole 64-bit word expansion cannot fit in one single witness element. Instead, they will be **split** into quarters of 64 bits each (corresponding to real 16 bits each), following this notation:
+
+$$\forall i \in \{0, 3\}:\quad split_i(sparse(X)) := sparse(X_{16i}^{16i+15})$$
```
joseandro, line 63: The interval `{0,3}` isn't being rendered correctly. It's not displaying the `{}` keys.

0008-zkvm-keccak.md
```diff
@@ -0,0 +1,791 @@
...
+
+Even though in each field element of ~254 bits long, one could fit up to 63 real bits with this encoding, each witness value will only store 16 such real bits. Instead, the expansion of the full 64 bit words could be represented by composition of each of the four 16-bit parts. But this will not be computed in the circuit, since the expansion would not fit entirely in the field. The way to perform this mapping is through a 16-bit lookup table (of $2^{16}$ entries, larger than that would be too costly). We call this the **sparse** lookup table. 
+
+Let $sparse(X)$ refer to a representation of $X$, where only the indices in the $4i$-th positions correspond to real bits, and the intermediate indices can contain auxiliary information. Connecting to the above, after performing a series of boolean operations to the $expand(X)$ one can obtain some $sparse(X)$, but both encode the same $X$. Concretely,
+
+$$\forall X\in\{b_i\}^\ell:\quad X \equiv \sum_{0}^{\ell-1} 2^i \cdot sparse(X)_{4i} $$
```
joseandro, line 59: The interval `{b_i}` isn't being rendered correctly. It's not displaying the `{}` keys.

0008-zkvm-keccak.md
```diff
@@ -0,0 +1,791 @@
...
+
+Here, we present the bitwise-sparse representation of bitstrings that will be used to obtain more efficient computations of boolean functions within our proof system. 
+
+Given that simple binary operations are very costly within a SNARK, the new idea consists on arithmetizing boolean operations as much as possible, becoming more efficient with finite field algebra. We introduce the _bitwise-sparse representation_ of bitstrings for this goal.
+
+#### Expansion
+
+With this notation, input bitstrings are expanded inserting three `0` bits in between "real" bits (hence the _sparsity_). The motivation for this, is to use intermediate bits to store auxiliary information such as carry bits, resulting from algebraic computations of boolean functions. That means, for each bit in the original bitstring, its expansion will contain four bits. The following expression represents such mapping:
+
+$$\forall X \in \{b_i\}^\ell, b\in\{0,1\}: expand(X) \to 0 \ 0 \ 0 \ b_{\ell-1} \ . \ . \ . \ 0 \ 0 \ 0 \ b_0$$
```
joseandro, line 49: The intervals `{b_i}` and `{0,3}` isn't being rendered correctly. It's not rendering the `{}` keys.

0008-zkvm-keccak.md
```diff
@@ -0,0 +1,791 @@
...
+\end{align*}
+$$
+
+Note that the sparse representation of full 64-bit words will never be computed for that reason. Instead, it will be compressed back again into their relevant bits form, resetting the intermediate bits between steps.
+
+In the implementation, each split is assumed to be one entry of arrays of length 4, so the representation of $split_i(X)$ will be `x[i]`. If the input in question is a 64-bit word, then each entry will contain 16 bits. If the input is in sparse representation, then each entry will contain 64 sparse bits.
+
+#### Compression
+
+Another set of functions will be defined to be used in this new design. For $i\in\{0,3\}$, let $shift_i(X)$ be the function that performs the AND operation $(\wedge)$ of the sparse representation of the word $2^{\ell}-1$ ("all-ones" word in binary), shifted $i$ bits to the left, with the input word $X$. More formally,
```
joseandro, line 80: The interval `{0,3}` isn't being rendered correctly. It's not displaying the `{}` keys.

0008-zkvm-keccak.md
```diff
@@ -0,0 +1,791 @@
...
+$$
+\begin{align*}
+sparse(X) \iff & shift_0(sparse(X)) + shift_1(sparse(X)) \\
++ & shift_2(sparse(X)) + shift_3(sparse(X))
+\end{align*}
+$$
+
+In order to check the real bits behind a sparse representation, one can perform a lookup with the result of the $shift_0$. Note that this table is equivalent to the sparse table presented above. Before, on input a 16-bit word (equivalent to the row index), the table contained the expanded representation. Here instead, on input the expansion (because $shift_0$ has intermediate bits set to zero), one can check the word. 
+
+The correctness of the remaining shifts $(i\in\{1,3\})$ should also be checked. For this, the witness value $aux_i := shift_i/2^i$ will be stored in the witness in such a way that each $aux_i$ can be looked up reusing the table for $shift_0$, and then the following constraint should hold:
```
joseandro, line 107: The interval `{0,3}` isn't being rendered correctly. It's not displaying the `{}` keys.

joseandro (changes requested):

0008-zkvm-keccak.md
querolita: That's probably the markdown/latex renderer in github. The one inside VSCode renders them well.

0008-zkvm-keccak.md
```diff
@@ -0,0 +1,815 @@
...
+\end{align*}
+$$
+
+The effect of each $shift_i$ is to null all but the $(4n+i)$-th bits of the expanded output. Meaning, they act as selectors of every other $(4n+i)$-th bit. That implies that one can rewrite the sparse representation of any word as:
+
+
+$$
+\begin{align*}
+sparse(X) \iff & shift_0(sparse(X)) + shift_1(sparse(X)) \\
++ & shift_2(sparse(X)) + shift_3(sparse(X))
```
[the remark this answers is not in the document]
querolita: Here, none of VSCode nor Git's renderer shows broken to me.

0008-zkvk-keccak.md
```diff
@@ -0,0 +1,665 @@
...
+        let remainder(x,y) = remainder_e(x,y)[0] + 2^16*remainder_e(x,y)[1] + 2^32*remainder_e(x,y)[2] + 2^48*remainder_e(x,y)[3]
+        let bound(x,y)     = bound_e(x,y)[0]     + 2^16*bound_e(x,y)[1]     + 2^32*bound_e(x,y)[2]     + 2^48*bound_e(x,y)[3]
+        let rotated(x,y)   = dense_rot_e(x,y)[0] + 2^16*dense_rot_e(x)[1] + 2^32*dense_rot_e(x,y)[2] + 2^48*dense_rot_e(x,y)[3]
+        constrain( word(x,y) * 2^(OFF(x,y)) - ( quotient(x) * 2^64 + remainder(x)) )
+        constrain( rotated(x,y) - (quotient(x,y) + remainder(x)) )
+        constrain( bound(x,y) - (quotient(x,y) + 2^64 - 2^(OFF(x,y)) )
+        constrain( state_b(y,2x+3y) - expand_rot_e(x,y) )
+```
+
+Note that the value `OFF(x,y)` will be a different constant for each index, according to the rotation table of Keccak.
```
mrmr1993, line 307: How/where is this encoded?

mrmr1993 (commented):
I don't see details here about how the data is fed in/out. I think we probably want to use the lookup argument (á la RAMLookup), but it would be good to solidify that design and call out the details

0008-zkvk-keccak.md
querolita: I thought it would be encoded directly onto the constraints themselves

querolita:
@mrmr1993 Not sure I know what you mean by "data is fed in/out ... RAMLookup"

Are you referring to how the message block goes from/to bytestrings to state ordering? If so, I added some notes in the appendix, as it is the same as in the old PoC.

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,778 @@
...
+### Constraints
+
+
+> For efficiency reasons, the state is assumed to be stored in expanded form, so that back-and-forth conversions do not need to take place repeatedly.
+
+The support for the Keccak hash function will require the following gate types:
+
+**PRECONDITIONS:** the message to be hashed should be padded with the $10^*1$ rule before 
+
+**NOTE:** we need to provide some mechanism to link blocks of the message to be hashed with the location in memory / public input, without relying on the permutation argument.
```
mrmr1993, line 169: This mechanism is missing.

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,778 @@
...
+
+        ```rust
+        constrain(root * old_state[i])
+        ```
+    
+    - **Pad mode** 
+
+        Padding with the $10^*1$ rule only happens once per hash. But unlike the root mode, padding happens only in the last absorb. That is because the input message is padded until reaching a length that is a multiple of 136 bytes, which always fully fits in one whole block. This means, in the case that the input message is $\leq 135$ bytes in length, the single absorb will run in both `root` and `pad` modes.
+
+        When in pad mode, we will have 136 selector flags (coefficient indices `[4..140)`) that will activate at most 136 positions corresponding to the padding. Additionally, each of the 136 coefficients corresponding to the bytes of the new block are set to either `0x00`, `0x01`, `0x80`, or `0x81`, which are all the possible combinations that a padding byte can take. In this mode, `absorb = 1, squeeze = 0`.
```
mrmr1993, line 245: These being 'coefficients' means that they're statically knowned and fixed in the circuit definition, right? I don't think that's the case if we don't statically know the length of hashes ahead of time (and I believe we don't).
It seems like we need some way of constraining these dynamically, in response to the preimages that are requested.

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,778 @@
...
+4. Testing requirements: 
+    * When the project is in a more advanced state, run integration tests with other parts of the system (MIPS ISA Implementation, and sys-call for blocks via Cannon).
+5. Testing resources: 
+    * Run the official [test vectors](https://keccak.team/archives.html).
+    * Obtain actual examples of blocks to hash, with different lengths.
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+This gate uses much longer lookup tables. Understand if the gains in the number of rows of Keccak compensate for the necessity of a $2^{16}$ lookup table.
```
mrmr1993, line 544: We want a 2^16 lookup table for the other instructions anyway; this is a non-concern.

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,778 @@
...
+The current Keccak PoC in SnarkyML was introduced to support Ethereum primitives in MINA zkApps. Due to this blockchain's design, the gadget needed to be compatible with Kimchi: a Plonk-like SNARK instantiated with Pasta curves for Pickles recursion and IPA commitents. This means that the design choices for that gadget were determined by some features of this proof system, such as: 15-column width witness, 7 permutable witness cells per row, access to the current and next rows, up to 4 lookups per row, access to 4-bit XOR lookup table, and less than $2^{16}$ rows. As a result, proving the Keccak hash of a message of 1 block length (up to 1080 bits) took ~15k rows.
+
+| Version  | Columns | Rows / block | Lookups / block | 
+|----------|---------|--------------|-----------------|
+| Old PoC  | 15      | $24\times(125+100+40+125+75+287.5+5)=18,180$ |$24\times(400+320+140+400+300+800+16)=57,024$ |   
+
+## Unresolved questions
+
+* During the review of this RFC:
+    * Resolve how many lookup tables are really needed (tradeoff between columns and rows)
```
mrmr1993, line 587: 3, as described, is worth it.

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,778 @@
...
+    * Resolve how to deal with input and output.
```
mrmr1993, line 588: TBD, based on Danny's RFC. However, we probably want to have some kind of 'keccak communication channel' for the syscall, where the syscall can do
```
lookup_aggreg += 1/(x + keccak_channel_table_id + value * joint_combiner^1)
```
and this chip handles reading them with
```
lookup_aggreg -= 1/(x + keccak_channel_table_id + value * joint_combiner^1)
```
so that the final lookup aggregation is unchanged iff the keccak chip has processed and proved all of the messages.

mrmr1993 (commented):
Mostly LGTM, but see comments

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,778 @@
...
+
+This RFC presents a design of a Kimchi chip for the Keccak hash function. It is optimized for the zkVM project, making use of our [generalized expression framework](./0007-generalized-expression-framework.md), and assuming a variant of Kimchi using a Nova-styled folding over BN128 and KZG commitments. The generalized expression framework will provide the ability to create arbitrary number of columns and lookups per row for custom gates. This allows more optimized approaches to address boolean SNARK-unfriendly relations.
+
+## Summary
+
+This document presents a design to deeply support the Keccak hash function for efficient pre-image oracle access within the proof system. To do so, it leverages a *bitwise sparse representation* to compute boolean operations (XORs, ANDs), and sequences of them.
+
+## Motivation
+
+One component of the Optimism stack system is the pre-image data oracle which is responsible for getting information about the EVM into the MIPS world. The OP stack currently uses the existing ledger which is built on top of Keccak hashes as a baseline.
```
dannywillems, line 11: for getting information about the chains*

0010-zkvm-keccak.md
querolita: But this seems to be a problem not only of padding, but of block width as well. Meaning, if the input happens to take more than 135 bytes, then the circuit will need to have more rows as well. So there's no "universal" keccak circuit at this point. So, if you need to know the length in order to setup the circuit with X-many absorbs, why would the padding be a problem? The only difference relies on a more accurate input length (number of absorbs only needs to know what multiple of the block length is upperbounded by the length of the input; whereas the padding technique inside the sponge itself requires to know the exact byte length, not just an interval).

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,788 @@
...
+
+This document presents a design to deeply support the Keccak hash function for efficient pre-image oracle access within the proof system. To do so, it leverages a *bitwise sparse representation* to compute boolean operations (XORs, ANDs), and sequences of them.
+
+## Motivation
+
+One component of the Optimism stack system is the pre-image data oracle which is responsible for getting information about the chains. The OP stack currently uses the existing ledger which is built on top of Keccak hashes as a baseline.
+
+Having a dedicated chip inside Kimchi to prove correct computation of Keccak hashes from pre-image data (blocks) will result in a more efficient design of the MIPS zkVM as a whole, meeting the performance goals of the [OP RFP](https://github.com/ethereum-optimism/ecosystem-contributions/issues/61#issuecomment-1611488039).
+
+As indicated in the PRD, any member of the OP ecosystem who wanted to prove the state transition between two blocks of an OP stack chain executing the Cannon program in the O(1) MIPS zkVM should complete within 2 days (before optimizations) to meet the latency estimates from the RFP.
```
dannywillems, line 15:
```suggestion
As indicated in the PRD, any member of the OP ecosystem who wanted to prove the state transition between two blocks of an OP stack chain executing the `op-program` program in the O(1) MIPS zkVM should complete within 2 days (before optimizations) to meet the latency estimates from the RFP. 
```

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,788 @@
...
+mask_0\left(sparse(X)\right) &:= 0 \ 0 \ 0 \ 1_{\ell-1} \ .\ .\ .\ 0 \ 0 \ 0 \ 1_{0} \ \wedge\ sparse(X) \equiv expand(X) \equiv reset(sparse(X))\\
+mask_1\left(sparse(X)\right) &:= 0 \ 0 \ 1_{\ell-1} \ 0 \ . \ . \ . \ 0 \ 0 \ 1_{0} \ 0 \ \wedge\ sparse(X) \\
+mask_2\left(sparse(X)\right) &:= 0 \ 1_{\ell-1} \ 0 \ 0 \ . \ . \ . \ 0 \ 1_{0} \ 0 \ 0 \ \wedge\ sparse(X) \\
+mask_3\left(sparse(X)\right) &:= \ 1_{\ell-1} \ 0 \ 0 \ 0 \ . \ . \ . \ 1_{0} \ 0 \ 0 \ 0 \wedge\ sparse(X) \\
+\end{align*}
+$$
+
+The effect of each $mask_i$ is to null all but the $(4n+i)$-th bits of the expanded output. Meaning, they act as selectors of every other $(4n+i)$-th bit. That implies that one can rewrite the sparse representation of any word as:
+
+$$
```
dannywillems, line 84: This does not render correctly, missing `&`?

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,788 @@
...
+
+Given expanded `left` and `right` inputs, the sparse representation of $XOR_{64}$ can be constrained in quarters of 16 bits, as:
+
+$$\forall i\in[0,3]: sparse(xor_i) = expand(left_i) + expand(right_i)$$
+
+Nonetheless, if no more than 15 boolean operations have been performed since the latest reset, one can use sparse representations of the left and right inputs as well.
+
+#### Reset
+
+After each step of Keccak, the sparse representation must be reset to avoid overflows of the intermediate bits. For $n \in [0, 3], let $shift_n = mask_n(sparse)/2^n$, the correct decomposition of a 64-bit word can be constrained as
```
dannywillems, line 115:
```suggestion
After each step of Keccak, the sparse representation must be reset to avoid overflows of the intermediate bits. For $n \in [0, 3]$, let $shift_n = mask_n(sparse)/2^n$, the correct decomposition of a 64-bit word can be constrained as
```

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,788 @@
...
+
+Here, we present the bitwise-sparse representation of bitstrings that will be used to obtain more efficient computations of boolean functions within our proof system. 
+
+Given that simple binary operations are very costly within a SNARK, the new idea consists on arithmetizing boolean operations as much as possible, becoming more efficient with finite field algebra. We introduce the _bitwise-sparse representation_ of bitstrings for this goal.
+
+#### Expansion
+
+With this notation, input bitstrings are expanded inserting three `0` bits in between "real" bits (hence the _sparsity_). The motivation for this, is to use intermediate bits to store auxiliary information such as carry bits, resulting from algebraic computations of boolean functions. That means, for each bit in the original bitstring, its expansion will contain four bits. The following expression represents such mapping:
+
+$$\forall X \in (b_i)^\ell, b\in[0,1]: expand(X) \to 0 \ 0 \ 0 \ b_{\ell-1} \ . \ . \ . \ 0 \ 0 \ 0 \ b_0$$
```
dannywillems, line 53: Opiniated comment: I liked the notation used in zcash to mention the size of the bitstring. They use $$\mathbb{B}^{64} \rightarrow \mathbb{B}^{256}$$ for a function mapping a 64 bit string to a 256 bit string. It helps to categorize the input/outputs by just reading the domain and image of the function.

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,788 @@
...
+
+> NOTE: the above means that after each step of the permutation, the expanded representation needs to be reset before starting the next step to ensure completeness of the encoding. 
+
+Even though in each field element of ~254 bits long, one could fit up to 63 real bits with this encoding, each witness value will only store 16 such real bits. Instead, the expansion of the full 64 bit words could be represented by composition of each of the four 16-bit parts. But this will not be computed in the circuit, since the expansion would not fit entirely in the field. The way to perform this mapping is through a 16-bit 2-column lookup table that we call `Reset` (of $2^{16}$ entries, anything significantly larger than that such as 32 bits would be too costly).
+
+Let $sparse(X)$ refer to a representation of $X$, where only the indices in the $4i$-th positions correspond to real bits, and the intermediate indices can contain auxiliary information. Connecting to the above, on input a word $X$, we can expand it in order to perform a series of boolean operations (up to 15) which results in some $sparse(X)$, encoding the same $X$. Then,
```
dannywillems, line 61: Nit: the notation `sparse(X)` lets the reader think it is a function $sparse : \mathbb{B}^{64} \rightarrow \mathbb{B}^{256}$, where, in reality, it is an element of $\mathbb{B}^{256}$.

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,788 @@
...
+##### AND
+
+Note that the AND of two expanded inputs corresponds to the $4n+1$-th positions after an expanded XOR (addition). Meaning that $shift_1 = mask_1/2$ gives the expanded representation of the conjugation operation.
+The AND operation can be constrained making use of XOR and Reset explained above. With no further constraints, XOR is used to obtain `left+right`, and then Reset provides the expanded result of AND in the `shift1` witnesses.
+
+#### Negation
+
+The NOT operation will be performed with a subtraction operation using the constant term $(2^{16}-1)$ for each 16-bit quarter, and only later it will be expanded. The mechanism to check that the input is at most 64 bits long, will rely on the composition of the sparse representation. Meaning, 64-bit words produce four 16-bit lookups, which implies that the real bits take no more than 64 bits. Given $x$ in reset form (meaning, not any sparse representation of $x$ but the initial expanded one with zero intermediate bits) the negation of one 64-bit word can be constrained as:
+
+$$\forall i \in[0, 3]: expand(not_i) = 0\text{x}1111111111111111 - expand(x_i)$$
```
dannywillems, line 132:
```suggestion
$$\forall i \in[0, 3]: expand(not_i) = 0\text{b}1111111111111111 - expand(x_i)$$
```

`0x` is for base 16, `0b` is for base 2.

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,788 @@
...
+\end{align*}
+$$
+
+### Basic toolbox
+
+#### XOR
+
+The XOR of 16-bit can be computed using addition of their expansions. The 16-bit output of the XOR will be located on the $4n$ positions, whereas the intermediate positions will contain any possible carry terms.
+
+Given expanded `left` and `right` inputs, the sparse representation of $XOR_{64}$ can be constrained in quarters of 16 bits, as:
```
dannywillems, line 107: To avoid ambiguity and ease the reading, I would define domains for each values, like if `left` is a 64 bits string, I would say
> Given an expanded input $\text{left} \in \mathbb{B}^{64}$

It is easier for the reader as it removes the work of checking (even if sometimes obvious) the domain of the operations.

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,788 @@
...
+
+$$\forall i\in[0,3]: sparse(xor_i) = expand(left_i) + expand(right_i)$$
+
+Nonetheless, if no more than 15 boolean operations have been performed since the latest reset, one can use sparse representations of the left and right inputs as well.
+
+#### Reset
+
+After each step of Keccak, the sparse representation must be reset to avoid overflows of the intermediate bits. For $n \in [0, 3]$, let $shift_n = mask_n(sparse)/2^n$, the correct decomposition of a 64-bit word can be constrained as
+
+$$\forall i \in [0,3]: sparse_i = shift0_i + 2\cdot shift1_i + 4\cdot shift2_i + 8\cdot shift3_i$$
```
dannywillems, line 117: I would explicitly say that $\forall i \in [0,3],  \text{sparse}_{i} \in \mathbb{B}^{16}$  and $\text{sparse} \in \mathbb{B}^{64}$

dannywillems: and maybe use $\text{shift}_{i, 0}$

and say it is $\text{shift}_{i, j} \in {0, 1}$

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,788 @@
...
+
+##### AND
+
+Note that the AND of two expanded inputs corresponds to the $4n+1$-th positions after an expanded XOR (addition). Meaning that $shift_1 = mask_1/2$ gives the expanded representation of the conjugation operation.
+The AND operation can be constrained making use of XOR and Reset explained above. With no further constraints, XOR is used to obtain `left+right`, and then Reset provides the expanded result of AND in the `shift1` witnesses.
+
+#### Negation
+
+The NOT operation will be performed with a subtraction operation using the constant term $(2^{16}-1)$ for each 16-bit quarter, and only later it will be expanded. The mechanism to check that the input is at most 64 bits long, will rely on the composition of the sparse representation. Meaning, 64-bit words produce four 16-bit lookups, which implies that the real bits take no more than 64 bits. Given $x$ in reset form (meaning, not any sparse representation of $x$ but the initial expanded one with zero intermediate bits) the negation of one 64-bit word can be constrained as:
```
dannywillems, line 130:
```suggestion
The NOT operation will be performed with a subtraction operation using the constant term $(2^{16}-1)$ for each 16-bit quarter, and only later it will be expanded. The mechanism to check that the input is at most 64 bits long, will rely on the composition of the sparse representation. Meaning, 64-bit words produce four 16-bit lookups, which implies that the real bits take no more than 64 bits. Given $x$ in canocical form (meaning, not any sparse representation of $x$ but the initial expanded one with zero intermediate bits) the negation of one 64-bit word can be constrained as:
```
? Maybe let's define `canonical` somewhere? I don't know if canonical is the best word in this case.

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,788 @@
...
+
+$$\forall i \in [0,3]: sparse_i = shift0_i + 2\cdot shift1_i + 4\cdot shift2_i + 8\cdot shift3_i$$
+
+together with a lookup for each `shift_i` (more about this later).
+
+Each `sparse[i]` will be the output of previous boolean operations. The `shift0[i]` corresponds to the clean expand of the word, and could be the input of some upcoming boolean operations. The `shift1[i]` will be used by the AND. The 64-bit word computed from the `dense[i]` will be used in rotation.
+
+##### AND
+
+Note that the AND of two expanded inputs corresponds to the $4n+1$-th positions after an expanded XOR (addition). Meaning that $shift_1 = mask_1/2$ gives the expanded representation of the conjugation operation.
```
dannywillems, line 125:
```suggestion
Note that the AND of two expanded inputs corresponds to the $4n+1$-th positions after an expanded XOR (addition). Meaning that $shift_1 = mask_1/2$ gives the expanded representation of the conjunction operation.
```

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,788 @@
...
+If the input was not previously expanded, the next step would be to expand all 25 words. Each word of 64 bits would have to be split into 4 parts of 16 real bits each. The expansion itself would be performed through the `Reset` lookup table. This step would require $4\times25=100$ lookups.
+
+There are two main modes: absorb and squeeze. Inside the absorb mode, there are two additional submodes called pad and root, which can as well happen at the same time. 
+
+
+- **Absorb mode**
+
+    In absorb mode, the first coefficient is set to 1: `absorb := coeff[0] = 1`, and the second coefficient is 0.
+
+    The gate takes a block as 68 chunks of 16 bits expanded, pads with zeros until reaching 100 chunks (1600 bits) to fit one whole state, and XORs it with the previous state.
```
dannywillems, line 200: Note (for me): `68 * 16 = 1088`, which is the rate of the sponge. We keep 512 bits in the capacity for 256 bits of security (sponge security is based on the capacity size).

dannywillems: (Note): the XOR is the same than with AO permutations. For a stat of `3`, we add after each permutation the block `[1, x, y]`.

dannywillems: Do we add the zeroes at the end or at the beginning of the chunks (also, are we always talking in little-endian or big-endian?).

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,788 @@
...
+> Assuming that we execute rotation starting form reset values, then one could perform the above operations directly on the expanded terms, since $4\cdot64-3=253$ encoded bits (removing the three leading zeros) would fit in the BN254 curve. In this scenario, rotations by $x$ bits would be mapped to rotations by an offset of $4\cdot x$ bits instead. Moreover, the constraints would need to be slightly different, to avoid overflows. In particular: `word = quotient * 2^(253-4*offset) + shift` and `output = quotient + shift * 2^(4*offset)`, and the output would already be in expanded format.  Nonetheless, in this case we would also need to perform bound checks, which would be much more costly due to the bitlengths. And in any case, each dense word should be expanded afterwards to be processed in later steps.
+
+### Constraints
+
+
+> For efficiency reasons, the state is assumed to be stored in expanded form, so that back-and-forth conversions do not need to take place repeatedly.
+
+The support for the Keccak hash function will require the following gate types:
+
+**PRECONDITIONS:** the message to be hashed should be padded with the $10^*1$ rule before
```
dannywillems, line 167: Is it possible to add a sentence (like in Appendix or redirect to a place describe it) explaining/formalizing the padding rule?

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,788 @@
...
+**PRECONDITIONS:** the message to be hashed should be padded with the $10^*1$ rule before 
+
+- `KeccakSponge`: performs formatting actions related to the Keccak sponge.
+- `KeccakRound`: performs one full round of the Keccak permutation function.
+
+The high-level layout of the gates follows:
+
+| `KeccakSponge` | [0...100) | [100...168) | [168...200) | [200...400) | [400...800) |
+| -------------- | --------- | ----------- | ----------- | ----------- | ----------- |
+| Curr           | old_state | new_block   | zeros       | bytes      | shifts      |
```
dannywillems, line 176: Do you confirm that new_block contains the non-expanded message, i.e. simply 68 bytes from the message?

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,788 @@
...
+Support for padding shall be provided. In the Keccak PoC, this step took place at the Snarky layer. It checks that the correct amount of bits in the $10^*1$ rule are added until reaching a multiple of 1088 bits, and then adds 512 more zero bits to each block to form a full state.
+
+If the input was not previously expanded, the next step would be to expand all 25 words. Each word of 64 bits would have to be split into 4 parts of 16 real bits each. The expansion itself would be performed through the `Reset` lookup table. This step would require $4\times25=100$ lookups.
+
+There are two main modes: absorb and squeeze. Inside the absorb mode, there are two additional submodes called pad and root, which can as well happen at the same time. 
+
+
+- **Absorb mode**
+
+    In absorb mode, the first coefficient is set to 1: `absorb := coeff[0] = 1`, and the second coefficient is 0.
```
dannywillems, line 198: I don't understand here: what is coefficient in this case? What is `1` (the one of the field?)

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,788 @@
...
+The support for the Keccak hash function will require the following gate types:
+
+**PRECONDITIONS:** the message to be hashed should be padded with the $10^*1$ rule before 
+
+- `KeccakSponge`: performs formatting actions related to the Keccak sponge.
+- `KeccakRound`: performs one full round of the Keccak permutation function.
+
+The high-level layout of the gates follows:
+
+| `KeccakSponge` | [0...100) | [100...168) | [168...200) | [200...400) | [400...800) |
```
dannywillems, line 174: What are the numbers? The column number, and therefore the element block will be 68 field elements encoding bytes in non-expanded form?

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,788 @@
...
+- **Absorb mode**
+
+    In absorb mode, the first coefficient is set to 1: `absorb := coeff[0] = 1`, and the second coefficient is 0.
+
+    The gate takes a block as 68 chunks of 16 bits expanded, pads with zeros until reaching 100 chunks (1600 bits) to fit one whole state, and XORs it with the previous state.
+
+    The zero padding is enforced performing the following 32 constraints:
+
+    ```rust
+    constrain(absorb * zeros[i])
```
dannywillems, line 205: absorb is a selector I suppose?

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,788 @@
...
+    ```
+
+    The XOR part of the gate uses $100$ constraints 
+
+    ```rust
+    constrain(absorb * (xor_state[i] - (old_state[i] + new_state[i])))
```
dannywillems, line 211: note (for me): and we have $xor_state[i] \in \mathbb{F}$

but it is in fact encoding a byte, i.e. the field value is below 2*16.

0010-zkvm-keccak.md
querolita: well, sparse of $X$ means $X$ is something in $\mathbb B^{64}$ and then outputs some element of $\mathbb B^{256}$ that encodes $X$ precisely. I mean $sparse$ is a function, but $sparse(X)$ is referring to the image of the function applied on $X$ so yeah, an element of $\mathbb B^{256}$

0010-zkvm-keccak.md
querolita: yeah I know, but it is indeed hexadecimal what we want to represent here. Actually, in the PRs, the way to compute expand easily is by reading in hexadecimal the binary representation of the input (this is inside polynomials/keccak/mod.rs, but also in the sparse lookup table generation code). 

If you were using binary above, then you would have overflows in the very next xor you would perform on top of the not. That's because all of the intermediate bits would be set to 1, which we have to avoid. 

In other words, the expand of the negation is equivalent to expanding $2^{64}$ and then subtracting the expand of X. 

`0x1111111111111111` is just the equivalent to writing $expand(2^{64})$

0010-zkvm-keccak.md
querolita: actually each sparse_i is in B64 and full sparse in B256. Okay, I see the need of becoming more specific as it seems error prone otherwise.

0010-zkvm-keccak.md
querolita: Just for clarity, 136 * 16 = 1088. Confusion might come from the fact that each sparse term codifies 2 bytes (16 bits).

querolita: > talking in little-endian or big-endian

The first byte (in memory, the one right after 0x.....) gets mapped to the least significant byte of the first word of the state. So that means, inside each word, it is little endian. But then when we obtain the hash, we invert this mapping so that the least significant byte of the first word of the state goes to the "first" byte of the hash.

0010-zkvm-keccak.md
querolita: coeff refers to the coefficients of the gate. Like in the generic gate, where they are used to encode the operation, here they encode modes of operation. The one is indeed a field element one.

0010-zkvm-keccak.md
querolita: The numbers are column indices of the witness layout. The first 68 positions encode the **expanded** 1088 bits of the preimage data. Meaning, each of those terms already encode 16 bits in its reset (canonical) form, taking 64 bits in the finite field element.

0010-zkvm-keccak.md
querolita: Absorb is a selector yes, as defined above as the first coefficient of the gate. It can equal either zero or one yes.

0010-zkvm-keccak.md
querolita: Each i-th position of those constraints encode 2 bytes (16 bits) as 64 bit field elements.

0010-zkvm-keccak.md
querolita: No, it contains the expanded message already. It contains the 68*16=1088 bits, since each cell encodes 2 bytes.

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,788 @@
...
+    * pack the bytestring given in chunks of 4 bytes.
+
+* Future work: 
+    * support SHA3 (NIST variant of Keccak), different output lengths, and different state widths;
+    * improve the lookup argument to reuse the `Reset` table also for single-column lookups. Perhaps this can work adding some logic such as performing lookups modulo a constant ($2^{64}$ in this case), or shifting right by a given number of positions ($64$ in this case);
+    * optimize the `KeccakRound` to remove redundant intermediate states (some preliminar ideas have been performed);
+    * optimize range checks or think of a different approach towards rotation.
+
+
+## Appendix
```
dannywillems, line 612: I think adding a link to the keccak.team website might be good. They have good resources and the pseudo code on the specs summary is clear.

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,790 @@
...
+
+#### Expansion
+
+With this notation, input bitstrings are expanded inserting three `0` bits in between "real" bits (hence the _sparsity_). The motivation for this, is to use intermediate bits to store auxiliary information such as carry bits, resulting from algebraic computations of boolean functions. That means, for each bit in the original bitstring, its expansion will contain four bits. The following expression represents such mapping:
+
+$$\forall X \in \mathbb B^\ell: expand(X) \to 0 \ 0 \ 0 \ b_{\ell-1} \ . \ . \ . \ 0 \ 0 \ 0 \ b_0$$
+
+The reason behind the choice of three empty intermediate bits in the sparse representation follows from the concrete use case of Keccak where no more than $2^4-1$ consecutive boolean operations are performed, and thus carries cannot overwrite the content of the bits to the leftmost bits.
+
+> NOTE: the above means that after each step of the permutation, the expanded representation needs to be reset before starting the next step to ensure completeness of the encoding.
```
dannywillems, line 57: Can you define "step of the permutation"?
Round? Functions (iota, chi, etc)?

0010-zkvm-keccak.md
dannywillems: I am highly in favor of having, for all elements, the set it belongs to. For functions, I wish we can write the domain and image. It is pretty confusing as we talk once about 16 bits, after that 64, after that 256, after that the field F, etc.

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,790 @@
...
+
+Given expanded `left`, `right` inputs both in $\mathbb B^{64}$, the sparse representation of $XOR_{64}$ can be constrained in quarters of 16 bits, as:
+
+$$\forall i\in[0,3]: sparse(xor_i) = expand(left_i) + expand(right_i)$$
+
+Nonetheless, if no more than 15 boolean operations have been performed since the latest reset, one can use sparse representations of the left and right inputs as well.
+
+#### Reset
+
+After each step of Keccak, the sparse representation must be reset to avoid overflows of the intermediate bits. For $n \in [0, 3]$, let $shift_n = mask_n(sparse)/2^n$, the correct decomposition of a 64-bit word can be constrained as
```
dannywillems, line 115: Step means round?

0010-zkvm-keccak.md
```diff
@@ -0,0 +1,790 @@
...
+0 = sparse(X) - (\ & reset_0(sparse(X)) + 2 \cdot shift_1(sparse(X)) \\
+& + 4 \cdot shift_2(sparse(X)) + 8 \cdot shift_3(sparse(X))\ )
+\end{align*}
+$$
+
+### Basic toolbox
+
+#### XOR
+
+The XOR of 16-bit can be computed using addition of their expansions. The 16-bit output of the XOR will be located on the $4n$ positions, whereas the intermediate positions will contain any possible carry terms.
```
dannywillems, line 105: NB: using the addition of the underlying field.

0010-zkvm-keccak.md
dannywillems: Edit: ok, it is the individual functions, i.e iota, rho, chi, theta, pi

PR #30, opened 2023-09-28
Syscalls

dannywillems:
Links to other RFC's are not correct as not merged yet.

### DIFF
--- a/0010-keccak-syscall.md
+++ b/0010-keccak-syscall.md
@@ -279,6 +279,9 @@ The variable pre-image key offset is reset to `0` everytime in this case, leavin
 place for optimisation in the MIPS code by avoiding calling `8` times the syscall
 if keys are not far away.
 
+At the end of the execution, the register `v0` must contain the number of bytes
+that have been written.
+
 <!--
 The pre-image key offset is used by the syscall `sysRead` on the file descriptor
 `fdPreimageRead` is used to know the chunk it is processing.
--- a/0010-keccak-syscall.md
+++ b/0010-keccak-syscall.md
@@ -194,10 +194,12 @@ chunk of the pre-image is read, and copy into the memory.
 The pre-image that must be read is determined by the `preimageKey`, and the actual
 chunk is given by the preimage offset variable.
 
-The number of bytes that must be read is in the register `a2`.
+The number of bytes that must be read is in the register `a2` and the address
+where the chunk as to be written is in the register `a1`.
 
 At the end of the call, the register `v0` must contain the number of bytes that
-have been read.
+have been read. Also, the preimage offset has to be incremented by the number of
+bytes that has been read.
 
 
 #### Hint response
--- a/0010-keccak-syscall.md
+++ b/0010-keccak-syscall.md
@@ -1,0 +1,354 @@
+# Keccak Syscalls
+
+## Summary
+
+A MIPS program is a succession of MIPS instructions. Some instructions are
+syscall instructions. Only a subset of syscalls have to be implemented for the
+OP MIPS zkVM. This document will describe how the syscalls have to be handled
+and how the state of the VM must be modified.
+A specific attention will be given to the pre-image reading and writing syscalls
+which requires some hashing using [Keccak][./keccak.md], therefore giving the
+name to this RFC.
+
+The program can request data from the outside world, but instead of using an
+RPC, the data will be loaded through an oracle, called the pre-image oracle. You
+can find a better description about it in the RFC [Client/Host
+integration](./0011-cli-host-integration.md).
+
+## Motivation
+
+We aim to implement the same syscalls in the MIPS zkVM than the current version
+that Cannon supports . The behavior must be the same. The VM must read and
+write to the same registers.
+
+## Detailed design
+
+Multiple general purpose registers will be used. We will give an alias for each
+of them in this document and we will refer to them by their name.
+
+<!-- https://www.dsi.unive.it/~gasparetto/materials/MIPS_Instruction_Set.pdf -->
+Here the mapping:
+- `zero` - register number 0
+- `v0` - register number 2
+- `v1` - register number 3
+- `a0` - register number 4
+- `a1` - register number 5
+- `a2` - register number 6
+- `a3` - register number 7
+
+This document will also suppose that the zkVM has some pre-defined interface to
+read and write the memory and the registers, and the details of the
+implementation and the underlying generated constraints will be omitted.
+Here provided the environment:
+<!-- TODO: add more while implementing -->
+- `set_register_value(idx, v)`: set the register number `idx` to the value `v`
+- `get_register_value(idx, buffer_v)`: copy the register number `idx` in the variable `buffer_v`
+- `set_halted(exit_code)`: constraint halting the program with the exit code `exit_code`.
+
+At the beginning of the instruction, the syscall number is in the register `v0`.
+
+The list of supported syscalls are:
+- `sysMmap`, mapped to value `4090`
+- `sysBrk`, mapped to value `4045`
+- `sysClone`, mapped to value `4120`
+- `sysExitGroup`, mapped to value `4246`
+- `sysRead`, mapped to value `4003`
+- `sysWrite`, mapped to value `4004`
+- `sysFcntl`, mapped to value `4055`
+
+Except in the case of the syscall `sysExitGroup`, at the end of the instruction,
+the instruction pointer must be increased by a word (32bits) as the next
+instruction pointer.
+
+For any other syscall value, the value `0` must be set into the registers `v0`
+and `a3`.
+
+An important note is that IO operations are truncated to `4` bytes, meaning that
+the register `v0` is at most `4`.
+Another important note is that addresses must be aligned at 4 bytes. IO syscalls
+**must** support unaligned addresses, i.e. the address `42` can be given even
+though it is not a multiple of `4`.
+Computing the effective address of an address `addr` can be done using the
+bitwise AND operator with `0xFFffFFfc`, which sets the latest two bits to `0`,
+i.e. takes the previous multiple of `4` which represents the aligned memory
+address. For instance, for the unaligned address `42`, the address `40` must be
+used while fetching the memory.
+
+Note that as the syscall instruction must be encoded in a circuit and its
+behavior depends on the value in the register `v0`, we must represent its effect
+using a polynomial constraining the value `v0`. We can easily do that using a
+vanishing polynomial on the values different than the one expected.
+For instance, let's suppose `v0` is `sysFnctl` and `sysFnctl` behavior is represented
+by the polynomial `P(v0, a0, a1, a3)`, we will use the polynomial:
+```
+P_syscall_fnctl(v0, a0, a1, a3) = \
+  (v0 - 4090) *
+  (v0 - 4045) *
+  (v0 - 4120) *
+  (v0 - 4246) *
+  (v0 - 4003) *
+  (v0 - 4004) *
+  P(v0, a0, a1, a3)
+```
+
+to activate the `P_syscall_fnctl` behavior when `v0 = 4055` and deactivate it
+when it has a different value. Reduction to polynomials of at most degree `2` can be
+done using [quadracization](./0008-kimchi-polynomial-quadracization.md).
+
+We suppose quadracization is going to be used, and therefore, we will not talk
+about the degree of the different polynomial constraints, and the detailled
+description will stay high level using the API described above, with some
+additional tweaks if necessary for a specific syscall behavior.
+
+Regarding the pre-oracle data, it will be loaded as part of the witness given to
+the prover.
+In the context of generating a proof using the zkVM, the program will first be
+run outside of the zkVM, and therefore we know which data and how much data will
+be requested by the program.
+A new vector of bytes can be added to the Witness structure, mimicking the work
+done for the memory, which will keep track of the data requested by the program.
+The [CLI](./cli.md) integration will take care of providing and populating
+correctly the Witness instance.
+
+Regarding the encoding, we must represent the preimage-key as
+a field element of the scalar field of the pairing-friendly curve BN128, which
+is 254 bits. The keys are 32 bytes long as it is the result of a KECCAK execution and
+therefore is too large to be saved in a field element.
+According to the [CLI RFC](./cli.md), the first byte of the hash must be dropped to be
+replaced by the request type, leaving therefore 248 bits to be encoded in a
+field element, which fits into an element of the scalar field of BN128.
+The first byte can be used to encode the table ID the pre-image data are going
+to be in.
+For the rest of the document, it will be supposed that only type 2 pre-image
+keys must be supported, i.e. the table ID will be considered to be a certain
+arbitrary constant.
+
+Regarding the the pre-image data, it is important to note that it can be of any
+size. The efficiency of a representation and encoding in a lookup table will
+depend on the sizes of the data and also its number, and therefore it is
+left to the kind of applications the VM is running.
+However, there are some optimisations that can be used.
+As the key is only 31 bytes, i.e. 248 bits, there are still 5 bits of data that
+can be used to encode an index of the data for a specific key. Therefore, for a
+given pre-image key, we can encode 2^5 * 31 bytes, i.e. 992 bytes. In terms of
+constraints, it will cost an additional constraint to mask the key and get the
+corresponding index. Using this encoding trick, and supposing the table can be
+of maximum size 2^16, we can encode 2048 different pre-image keys in one table.
+
+In addition to this technique, a wider table can also be used to encode bigger data.
+
+<!-- TODO: a word on the pre-image key as we can only read 4 bytes by 4 bytes,
+     we need to save it somewhere -->
+
+### `sysMmap`
+
+<!-- DONE -->
+
+This can be ignored at the moment and implemented as a no-op.
+For consistency with Cannon, `v0` must be set to the value of the register `a0`
+and `a3` must be set to `0`.
+
+### `sysBrk`
+
+<!-- DONE -->
+
+The register `v0` must be set to `0x400000` and the register `a3` to `0`.
+
+### `sysClone`
+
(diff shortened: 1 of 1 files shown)


0009-folding.md
```diff
@@ -31,15 +31,15 @@ single one.
 ### Folding in kimchi
 
 For kimchi, the witness is the same witness we already have: a list of columns/polynomials. And the instance is list of the
-commitments to those polynomial, plus the public inputs.  
+commitments to those polynomials, plus the public inputs.
```
mrmr1993, line 34: Please make a separate PR if you want to tweak grammar in an existing RFC.

mrmr1993 (commented):
This appears to still be a draft. Also, see comment.

dannywillems:
This comment will be edited, and it is only to keep track of stats about the data. Might be useless.
```
- op-program-db-sepolia-2023-10-18-17-17-55

L1_HEAD=0xc58ce712ce933238b18add449ff05967ea2df66144b4f5441be305df3bc6ce87
L2_HEAD=0xa48a6758889af82824116aabd3ab1f22fe6be91b0a64a01165ec7910939cf45f
L2_BLOCK_NUMBER=2919120
STARTING_OUTPUT_ROOT=0xc5aecf6db3d8016a47c2acd7b7837d537d0dea8c359fd6f2c7d636f5159f9c19
L2_CLAIM=0xc5aecf6db3d8016a47c2acd7b7837d537d0dea8c359fd6f2c7d636f5159f9c19

$ find . -size -992c | wc -l
2039
--> can be encoded in one table, with one chunk

$ ls -l | wc -l
4107
```

dannywillems:
Important note: https://github.com/ethereum-optimism/optimism/blob/a2cc2f41fc7cf270b3e44471a7a4e4931ccaf121/cannon/mipsevm/open_mips_tests/test/oracle.asm#L72

> # reading the pre-image stream at EOF should have no effect

Not checked yet if the implementation handles this.

dannywillems:
(draft)
Idea to integrate the keccak gate:
First, keccak is a sponge based hash function. It uses a state of 1600 bits, 512 bits for the capacity (for 256 bits of security) and a absorbion rate of 1088 bits, i.e. 136 bytes.
In circuits, keccak uses the columns from 100 to 168.
bits, i.e. 136 bytes.

When we do a syscall `sysWrite` with the fd `premageWrite`, we have to reset the keccak state to the initial state, as we do for the preimage offset. We can see `sysWrite preimageWrite` as a reset of the preimage communication channel.
Q: what do we do if the keccak state is not empty, i.e. we have been reading data and we haven't run keccak and verify the hash?
It can be the moment we `finalize` the preimage reading. In a sense, when we call preimageWrite, we are preparing the VM to read new preimage data, and it can be the trigger that we read all previous data we had to.


Let's start reading the preimage with sysRead. Every bytes we read, we can copy it in the appropriate columns. The preimage offset will give the number of 

We can run the permutation when we have read 136 bytes with syscall.

0010-keccak-syscall.md
```diff
@@ -0,0 +1,402 @@
...
+  (v0 - 4004) *
+  P(v0, a0, a1, a3)
+```
+
+to activate the `P_syscall_fnctl` behavior when `v0 = 4055` and deactivate it
+when it has a different value. Reduction to polynomials of at most degree `2` can be
+done using [quadracization](./0008-kimchi-polynomial-quadracization.md).
+
+We suppose quadracization is going to be used, and therefore, we will not talk
+about the degree of the different polynomial constraints, and the detailled
```
rbonichon, line 99: typo
```suggestion
about the degree of the different polynomial constraints, and the detailed
```

0010-keccak-syscall.md
```diff
@@ -0,0 +1,402 @@
...
+
+Regarding the encoding, we must represent the preimage-key as
+a field element of the scalar field of the pairing-friendly curve BN128, which
+is 254 bits. The keys are 32 bytes long as it is the result of a KECCAK execution and
+therefore is too large to be saved in a field element.
+According to the [CLI RFC](./cli.md), the first byte of the hash must be dropped to be
+replaced by the request type, leaving therefore 248 bits to be encoded in a
+field element, which fits into an element of the scalar field of BN128.
+The first byte can be used to encode the table ID the pre-image data are going
+to be in.
```
rbonichon, line 121: Is that the first byte of the remaining 248 bits or an internal use of the first byte of the 256 bits value, since it disappears anyway in communication?

0010-keccak-syscall.md
```diff
@@ -0,0 +1,402 @@
...
+For the rest of the document, it will be supposed that only type 2 pre-image
+keys must be supported, i.e. the table ID will be considered to be a certain
+arbitrary constant.
+
+Regarding the the pre-image data, it is important to note that it can be of any
+size. The efficiency of a representation and encoding in a lookup table will
+depend on the sizes of the data and also its number, and therefore it is
```
rbonichon, line 128: > depend on the sizes of the data and also its number, and therefore it is

Does "number" have to be read as "quantity"? Is "sizes" really plural?

0010-keccak-syscall.md
```diff
@@ -0,0 +1,402 @@
...
+left to the kind of applications the VM is running.
+However, there are some optimisations that can be used.
+As the key is only 31 bytes, i.e. 248 bits, there are still 5 bits of data that
```
rbonichon, line 131: I don't follow where the 5 bits come from. Is that using the space left in the byte encoding the key type which we assume to be 0x02?

0010-keccak-syscall.md
```diff
@@ -0,0 +1,402 @@
...
+
+The preimage data length is located in the 8 first addresses, encoded on 64
+bits in big-endian.
+
+A difference with Cannon is we verify the preimage data is the actual data the
+user has requested previously. To achieve this, we will use the optimised
+keccak gate described in [Keccak RFC](./keccak.md).
+
+The keccak sponge state is 136 bytes, and is located in the columns `[100,
+168(`. While reading the preimage data, we can fill at the same time the sponge
```
rbonichon, line 212: > `[100, 168(`. 

This should probably read either `[100, 168[` or `[100, 168)`

0010-keccak-syscall.md
dannywillems: > Is that the first byte of the remaining 248 bits
Yes. Field elements are on 254 bits, therefore we still have 5 bits (the sixth bit can not be totally used as the modulo is slightly smaller than 2**254.

(Note that this remark won't be useful I think as we use another kind of lookup ("RAMlookup") that I had in mind when writing this: I will remove it later).

0010-keccak-syscall.md
dannywillems: > Does "number" have to be read as "quantity"?

Yes.

> Is "sizes" really plural?

I used plural as the preimages do not have all the same size.

0010-keccak-syscall.md
dannywillems: Yes

0010-keccak-syscall.md
```diff
@@ -0,0 +1,402 @@
...
```
dannywillems, line 212:
```suggestion
168[`. While reading the preimage data, we can fill at the same time the sponge
```

0010-keccak-syscall.md
dannywillems: Fixed. Thanks.

0010-keccak-syscall.md
```diff
@@ -0,0 +1,402 @@
...
+Another important note is that addresses must be aligned at 4 bytes. IO syscalls
+**must** support unaligned addresses, i.e. the address `42` can be given even
+though it is not a multiple of `4`.
+Computing the effective address of an address `addr` can be done using the
+bitwise AND operator with `0xFFffFFfc`, which sets the latest two bits to `0`,
+i.e. takes the previous multiple of `4` which represents the aligned memory
+address. For instance, for the unaligned address `42`, the address `40` must be
+used while fetching the memory.
+
+Note that as the syscall instruction must be encoded in a circuit and its
```
querolita, line 77: Mmmm so this sentence means that there will be a selector for each instruction type. So there's one for the syscall instruction. And when that selector is true, then the subtype of syscall being executed in particular is given by the content of the v0 register?

0010-keccak-syscall.md
```diff
@@ -0,0 +1,402 @@
...
+
+We suppose quadracization is going to be used, and therefore, we will not talk
+about the degree of the different polynomial constraints, and the detailed
+description will stay high level using the API described above, with some
+additional tweaks if necessary for a specific syscall behavior.
+
+Regarding the pre-oracle data, it will be loaded as part of the witness given to
+the prover.
+In the context of generating a proof using the zkVM, the program will first be
+run outside of the zkVM, and therefore we know which data and how much data will
```
querolita, line 106: Okay this is key. So this means, there is an interpreter of this program that extracts all needed data for this particular instance of the op program being proven validity about. So the code will generate some kind of struct storing all needed values for registers, instructions, etc, after "executing" that code, so that the circuit can be built accordingly?

If so, it seems like that execution will also give us the desired length of the keccak preimage, so the gadget can be called with the right amount and set the padding coefficients accordingly in the circuit. This seems like a good thing!

0010-keccak-syscall.md
```diff
@@ -0,0 +1,402 @@
...
+be requested by the program.
+A new vector of bytes can be added to the Witness structure, mimicking the work
+done for the memory, which will keep track of the data requested by the program.
+The [CLI](./cli.md) integration will take care of providing and populating
```
querolita, line 110: Oh, so it is the CLI (Richard's part?) which runs the MIPS program and creates such witness struct with all required data? I see how everything connects :D

0010-keccak-syscall.md
```diff
@@ -0,0 +1,402 @@
...
+correctly the Witness instance.
+
+Regarding the encoding, we must represent the preimage-key as
+a field element of the scalar field of the pairing-friendly curve BN128, which
+is 254 bits. The keys are 32 bytes long as it is the result of a KECCAK execution and
+therefore is too large to be saved in a field element.
+According to the [CLI RFC](./cli.md), the first byte of the hash must be dropped to be
+replaced by the request type, leaving therefore 248 bits to be encoded in a
```
querolita, line 118: This is interesting. I wonder if I should do anything differently in the "output" of the sponge to facilitate this "composition" of 31 bits, so you don't have to. This will just add 1 more constraint and 1 more witness cell on my side.

0010-keccak-syscall.md
```diff
@@ -0,0 +1,402 @@
...
+field element, which fits into an element of the scalar field of BN128.
+The first byte can be used to encode the table ID the pre-image data are going
+to be in.
+For the rest of the document, it will be supposed that only type 2 pre-image
+keys must be supported, i.e. the table ID will be considered to be a certain
+arbitrary constant.
+
+Regarding the the pre-image data, it is important to note that it can be of any
```
querolita, line 126: According to what we investigated the other day, blocks an at most take 1MB, even if they reserve space for up to 8 bytes to store the length of the preimage (2^64 length?!).

querolita: Now that I am recalling that call we had with Mofi, I believe he said 4 bytes instead, so up to 2^32 for length. But that's like 512MB instead of just 1...

0010-keccak-syscall.md
```diff
@@ -0,0 +1,402 @@
...
+size. The efficiency of a representation and encoding in a lookup table will
+depend on the sizes of the data and also its number, and therefore it is
+left to the kind of applications the VM is running.
+However, there are some optimisations that can be used.
+As the key is only 31 bytes, i.e. 248 bits, there are still 5 bits of data that
+can be used to encode an index of the data for a specific key. Therefore, for a
+given pre-image key, we can encode 2^5 * 31 bytes, i.e. 992 bytes. In terms of
+constraints, it will cost an additional constraint to mask the key and get the
+corresponding index. Using this encoding trick, and supposing the table can be
+of maximum size 2^16, we can encode 2048 different pre-image keys in one table.
```
querolita, line 136: I don't get why we want to encode preimage keys in a table 😞

0010-keccak-syscall.md
```diff
@@ -0,0 +1,402 @@
...
+bytes that has been read.
+
+The preimage data length is located in the 8 first addresses, encoded on 64
+bits in big-endian.
+
+A difference with Cannon is we verify the preimage data is the actual data the
+user has requested previously. To achieve this, we will use the optimised
+keccak gate described in [Keccak RFC](./keccak.md).
+
+The keccak sponge state is 136 bytes, and is located in the columns `[100,
```
querolita, line 211: Actually it is 200 bytes (1600 bits). It's just that at the beginning of the absorb, the last 64 bytes are zeros (meaning, they don't correspond to preimage data).

0010-keccak-syscall.md
```diff
@@ -0,0 +1,402 @@
...
+168[`. While reading the preimage data, we can fill at the same time the sponge
```
querolita, line 212: Mmmm yeah, basically the CLI (or maybe somewhere else) will store the preimage data and the keccak gadget will read from that variable (so far it is a biguint) and creates the corresponding witness terms for the keccak rows.

0010-keccak-syscall.md
```diff
@@ -0,0 +1,402 @@
...
+state. When 136 bytes have been read, a keccak permutation is executed.
+If we have finished reading the data, we run the permutation. In the case the total
```
querolita, line 214: Yeah, I mean you cannot execute the witness generation code of keccak until you have the whole block (at least). That can give you a partial witness in any case. If you want the whole thing, then you need to wait until all of the preimage has been read.

0010-keccak-syscall.md
```diff
@@ -0,0 +1,402 @@
...
+length is a multiple of 136, an additional permutation is executed
```
querolita, line 215: This is not super accurate but okay I know what you mean. Basically, the preimage data needs to be padded until it has a length that is a multiple of 1088 bits. That means you might have to run another absorb step if you end up with more than one block of 1088 bits each.

0010-keccak-syscall.md
```diff
@@ -0,0 +1,402 @@
...
+with the expected padding.
+
+Reading additional bytes after the EOF must not have any effect and must be
+allowed. This can be achieved by activating the "preimage read gate" iff the
+different between the number of bytes that have to be read and the preimage
```
querolita, line 220:
```suggestion
difference between the number of bytes that have to be read and the preimage
```

0010-keccak-syscall.md
```diff
@@ -0,0 +1,402 @@
...
+offset is not zero. In the case we arrived at the end of the file, the preimage
+offset must not be increased and the number of bytes read must be set `0`.
+
+After we read the data and run the permutation a last time as described above,
+we must verify that the preimage key corresponds to the output of sponge.
```
querolita, line 225: Mmmm trying to connect the dots here...

So this means, when you were reading preimage keys before, you wanted to store them in a table and I didn't understand why. Is this supposed to be the reason? After running keccak, you want to check that the log in the table for this file corresponds (without the first byte) to the output of the keccak gadget? Is this the motivation behind the "communication channel"?

If so, is it a runtime table where you write terms and right after when executing keccak, you remove from it? With the +/- technique?

0010-keccak-syscall.md
```diff
@@ -0,0 +1,402 @@
...
+To achieve this, we compare the 31 bytes first bytes as the first byte is
+dropped for the preimage key type. The output of the keccak sponge is described
```
querolita, line 227: yeah, so basically where and how do you do that? You're saying you will store the 31 bits of the key in that table and check that the (chopped) output of the keccak appears in that table? But then how do you link this one to a particular entry of the table? I mean, a normal lookup contains many terms in the table, and it solely gives you the guarantee that some term IS in the table. But doesn't tell you which entry. Here you not only want to check that you saw that (chopped) hash before as a file name, but you want to link it with the one you're just reading. How is this chain of connexions supposed to take place?

0010-keccak-syscall.md
```diff
@@ -0,0 +1,402 @@
...
+output. It can therefore be ignored.
+The number of bytes, i.e. the value in `a2`, must be copied into `v0` and
+the value `0` must be set in `a3`.
+
+#### Pre-image Oracle
+
+<!-- DONE -->
+
+The file descriptor `fdPreimageWrite` (value `6`) is used to read a pre-image
+key from the memory.
```
querolita, line 293: when is this memory loaded with data so you can later read from it? is it the CLI?

querolita (commented):
Excellent job! Left some comments to make sure I understand the approach.

0010-keccak-syscall.md
rbonichon: Would you mind making that clearer?

0010-keccak-syscall.md
rbonichon: Ok for the use of size**s**. The phrasing for "number" could be clearer, I think, but if it's clear to everyone else, I don't mind keeping it that way.

0010-keccak-syscall.md
rbonichon: Are we waiting for the rewrite?

0010-keccak-syscall.md
```diff
@@ -0,0 +1,402 @@
...
+An important note is that IO operations are truncated to `4` bytes, meaning that
+the register `v0` is at most `4`.
+Another important note is that addresses must be aligned at 4 bytes. IO syscalls
+**must** support unaligned addresses, i.e. the address `42` can be given even
+though it is not a multiple of `4`.
+Computing the effective address of an address `addr` can be done using the
+bitwise AND operator with `0xFFffFFfc`, which sets the latest two bits to `0`,
+i.e. takes the previous multiple of `4` which represents the aligned memory
+address. For instance, for the unaligned address `42`, the address `40` must be
+used while fetching the memory.
```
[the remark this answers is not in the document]
rbonichon: ^^

PR #35, opened 2023-10-16
[0011] zkApp-CLI enhancements for local lightnet management.

shimkiv:
Closes https://github.com/o1-labs/quality-and-testing/issues/139


0010-zkapp-cli-lightnet-management.md
```diff
@@ -0,0 +1,100 @@
...
+- **Output**:
+  - Anticipated network properties (as per [Docker image description](https://hub.docker.com/r/o1labs/mina-local-network));
+  - Mina Daemon GraphQL endpoint for communication: [http://localhost:8080/graphql](http://localhost:8080/graphql)
+  - PostgreSQL connection string: `postgresql://postgres:postgres@localhost:5432/archive`;
+  - Mina Daemon and Archive Node logs path;
+  - [o1js configuration](https://github.com/o1-labs/o1js/blob/ccf50e0b58190d9700c6dda4fafee3e62e270131/src/examples/zkapps/hello_world/run_live.ts#L18) code snippet.
+
+#### Stop lightnet network
+
+- **Command**: `zk lightnet stop`
```
MartinMinkov, line 48: From my understanding, docker containers will leave behind logs on the user's system. This could introduce a ton of bloat where logs are being kept from each run. Do you think we should wipe the logs out when tearing down the network or have some subcommand to do some house cleaning? Or should we just let the users worry about that, wdyt?

0010-zkapp-cli-lightnet-management.md
```diff
@@ -0,0 +1,100 @@
...
+#### Start lightnet network
+
+- **Command**: `zk lightnet start <options>`
+- **Function**: Checks the env., starts the local lightnet Docker container with reasonable default configuration options and waits until the network sync. Repeating the command while network is up and running should not start a new container, but rather check the status of the existing one.
+- **Configuration options**:
+  - --mode: `single-node` or `multi-node` (default: `single-node`)
+  - --proof-level: `none` or `full` (default: `none`)
+  - --archive-node: `true` or `false` (default: `true`)
+  - --mina-branch: `rampup`, `berkeley` or `develop` (default: `rampup`)
+  - --dune-profile: `lightnet` or `devnet` (default: `lightnet`)
```
MartinMinkov, line 38: When would someone use `devnet` for the dune profile? Is this something we need to expose? Are there invalid `mina-branch` and `dune-profile` combinations to look out for?

0010-zkapp-cli-lightnet-management.md
```diff
@@ -0,0 +1,100 @@
...
+
+- **Command**: `zk lightnet`
+- **Output**: Help message with description of available sub-commands and their options.
+
+#### Start lightnet network
+
+- **Command**: `zk lightnet start <options>`
+- **Function**: Checks the env., starts the local lightnet Docker container with reasonable default configuration options and waits until the network sync. Repeating the command while network is up and running should not start a new container, but rather check the status of the existing one.
+- **Configuration options**:
+  - --mode: `single-node` or `multi-node` (default: `single-node`)
```
MartinMinkov, line 34: How many nodes will be multi-node?

0010-zkapp-cli-lightnet-management.md
```diff
@@ -0,0 +1,100 @@
...
+#### Stop lightnet network
+
+- **Command**: `zk lightnet stop`
+  - In future we might want to extend this command with additional options to preserve the network state between containers lifecycle.
+- **Function**: Destroys the running container. Repeating the command while network is down should not throw an error but rather output a message that the network is already down.
+- **Output**: Maybe blockchain state (blocks height, slots, epoch, txns processed during the session, etc.) at the time of command issuing.
+
+#### Handle logs
+
+- **Command**: `zk lightnet logs <sub-command>`
```
MartinMinkov, line 55: Do you think it would be helpful to specify logs of specific nodes? Maybe only fetching logs from the single node or archive node? Do you see any value in that?

0010-zkapp-cli-lightnet-management.md
```diff
@@ -0,0 +1,100 @@
...
+- **Sub-commands**:
+  - **view**: `less` or `tail` the log files (Mina Daemon or Archive Node). OS agnostic.
+  - **save**: Saves the log files to default location and outputs the resulting path.
```
MartinMinkov, line 58: What will the default location be?

0010-zkapp-cli-lightnet-management.md
```diff
@@ -0,0 +1,100 @@
...
+## Detailed design
+
+### zkapp CLI Interface Enhancements
+
+Under the primary command `zk`, the new `lightnet` sub-commands will be introduced:
+
+#### Help
+
+- **Command**: `zk lightnet`
+- **Output**: Help message with description of available sub-commands and their options.
```
ymekuria, line 27: I like this pattern to list the help message with sub commands after entering `zk lightnet`. We should also add `zk lightnet ` and the subcommands to the help message a user can invoke when entering 'zk' or 'zk --help'. It can look something like this.

Usage: zk <command> [options]
```

Commands:
  zk project [name]  Create a new project                     [aliases: proj, p]
  zk file [name]     Create a file and generate the corresponding test file
                                                                    [aliases: f]
  zk config          Add a new deploy alias
  zk deploy [alias]  Deploy or redeploy a zkApp
  zk example [name]  Create an example project                      [aliases: e]
  zk system          Show system info                          [aliases: sys, s]
  zk lightnet [sub-command]
        start <options>       Start  a light network              [aliases: tbd]
           List start options
        stop <options>        Stop  a light network               [aliases: tbd]
            List stop options
        logs  <sub-command>       Display network logs            [aliases: tbd]
            List logs sub-commands
        explorer     Start  a explorer                            [aliases: tbd]
Options:
  -h, --help     Show help                                             [boolean]
  -v, --version  Show version number                                   [boolean]

```

MartinMinkov:
Another question I have is related to fault tolerance. How should we plan around node failures? If a node fails, is there any restart policies that happen automatically? If not, could we detect node errors and preemptively fix them? Should we notify users iof these node failures right away? What strategy would you like to see?

0010-zkapp-cli-lightnet-management.md
```diff
@@ -0,0 +1,100 @@
...
+  - --proof-level: `none` or `full` (default: `none`)
+  - --archive-node: `true` or `false` (default: `true`)
+  - --mina-branch: `rampup`, `berkeley` or `develop` (default: `rampup`)
+  - --dune-profile: `lightnet` or `devnet` (default: `lightnet`)
+- **Output**:
+  - Anticipated network properties (as per [Docker image description](https://hub.docker.com/r/o1labs/mina-local-network));
+  - Mina Daemon GraphQL endpoint for communication: [http://localhost:8080/graphql](http://localhost:8080/graphql)
+  - PostgreSQL connection string: `postgresql://postgres:postgres@localhost:5432/archive`;
+  - Mina Daemon and Archive Node logs path;
+  - [o1js configuration](https://github.com/o1-labs/o1js/blob/ccf50e0b58190d9700c6dda4fafee3e62e270131/src/examples/zkapps/hello_world/run_live.ts#L18) code snippet.
```
ymekuria, line 44: These are great outputs! We need to supplement this with docs walking developers through how to connect and use the light network in their workflow. We should have some text output in the CLI with all the information above, next immediate steps, and maybe a link to docs when they are completed.

0010-zkapp-cli-lightnet-management.md
```diff
@@ -0,0 +1,100 @@
...
+
+#### Stop lightnet network
+
+- **Command**: `zk lightnet stop`
+  - In future we might want to extend this command with additional options to preserve the network state between containers lifecycle.
```
ymekuria, line 49: +1 I think this can be a useful feature but we shouldn't focus on it for the first version.

0010-zkapp-cli-lightnet-management.md
```diff
@@ -0,0 +1,100 @@
...
+
+#### Help
+
+- **Command**: `zk lightnet`
+- **Output**: Help message with description of available sub-commands and their options.
+
+#### Start lightnet network
+
+- **Command**: `zk lightnet start <options>`
+- **Function**: Checks the env., starts the local lightnet Docker container with reasonable default configuration options and waits until the network sync. Repeating the command while network is up and running should not start a new container, but rather check the status of the existing one.
```
ymekuria, line 32: What will be checked in the env. ? Is the env. configured by the options entered in `zk lightnet start <options>`  or from somewhere else?

0010-zkapp-cli-lightnet-management.md
ymekuria: I have this same question. We could reuse the similar location we store cached feepayer information.

0010-zkapp-cli-lightnet-management.md
```diff
@@ -0,0 +1,100 @@
...
+- **Testing goals and objectives**:
+  - To validate the reliability of the enhanced zkapp CLI tool.
+- **Testing approach**:
+  - Unit tests for the new `lightnet` subcommands.
+  - Integration tests for the new `lightnet` subcommands.
+- **Testing scope**:
+  - Testing all lightnet commands and their possible variations.
+  - Addressing potential error scenarios and recovery mechanisms.
+- **Testing requirements**:
+  - Test in various local setups to simulate different development environments.
```
ymekuria, line 90: In the future it may be useful to test interacting with the light network from a browser UI.

ymekuria (approved):

ymekuria (approved):
This looks great and will be very useful for developers. I had some comments, questions, and suggestions.

0010-zkapp-cli-lightnet-management.md
```diff
@@ -0,0 +1,100 @@
...
+
+- **Command**: `zk lightnet account <sub-command>`
+- **Sub-commands**:
+  - **fetch**: Fetches available account from the container's pool.
+  - **release**: Releases "used" account back to the pool.
+- **Note**: Corresponding [o1js API](https://github.com/o1-labs/o1js/pull/1167) can be used.
+
+#### Launch the blockchain explorer
+
+- **Command**: `zk lightnet explorer`
```
garwalsh, line 70: @shimkiv just confirming that this functionality will be added once everything else here has been released, and after archive node API. Correct?

0010-zkapp-cli-lightnet-management.md
garwalsh: Agreed that this is a future requirement but I would love to see this happen at some point.

0010-zkapp-cli-lightnet-management.md
shimkiv: Correct.

0010-zkapp-cli-lightnet-management.md
shimkiv: Thank you @ymekuria, that was the idea.

0010-zkapp-cli-lightnet-management.md
shimkiv: Thanks @MartinMinkov,

> docker containers will leave behind logs on the user's system
Which logs you're referring here to?

As it is stated in description to `stop` command:
> In future we might want to extend this command with additional options to preserve the network state between containers lifecycle.
Which essentially means that for this initial version we will wipe everything (containers will start with `--rm` option and we will clean up other leftovers is any). RFC is updated to clearly state about the cleaning up procedure.

0010-zkapp-cli-lightnet-management.md
shimkiv: > When would someone use devnet for the dune profile? 
In case when one will need the network properties closer to those in use for public networks. Use `multi-node` as network mode, `full` as proof level, `devnet` as Dune profile and you will get container that runs small network of predefined participants with constants and properties closer to the real world.

> Is this something we need to expose?
Yes, for fine tuning I think we need it.

> Are there invalid mina-branch and dune-profile combinations to look out for?
No, every major Mina branch that we publish [docker image for](https://hub.docker.com/r/o1labs/mina-local-network/tags) is built twice: with `lightnet` and `develop` Dune profiles. This results in 2 images per Mina branch. Image tags anatomy is also described in Docker Hub [repo description](https://hub.docker.com/r/o1labs/mina-local-network).

Do you think we need to document something in more details?

0010-zkapp-cli-lightnet-management.md
shimkiv: The Docker Hub [repo description](https://hub.docker.com/r/o1labs/mina-local-network) has `Multi-Node ports reference` section which gives the network topology overview by participants exposing their ports. Maybe we need to describe it in more details on Docker Hub page.
Yes, as of now it is static/predefined network topology enough to keep network alive. Reason to have it I've described in discussion above.

0010-zkapp-cli-lightnet-management.md
shimkiv: I think the majority will start default networks (which is `single node network`) and because we operate on user's local machines there probably won't be much benefits of being that specific. For both modes I was thinking to just compress all logs available, save the archive and output the created file path (when user decides to save logs).
Viewing logs on another hand should have mentioned option to see logs of specific network participant.
Maybe we should not even expose `mode` and always run networks in `single-node` mode?

0010-zkapp-cli-lightnet-management.md
shimkiv: Yes I was thinking about what @ymekuria suggested.

0010-zkapp-cli-lightnet-management.md
shimkiv: Agree!

0010-zkapp-cli-lightnet-management.md
shimkiv: Here by "checks the env." I was only thinking about:
- check that docker engine installed and accessible by user
- check internet connectivity to make it possible download the image
- check that system has enough space to unpack the image and start the container

0010-zkapp-cli-lightnet-management.md
shimkiv: To be fair I need to say that the `light network` is already in use for `zkapp-cli` and `o1js` E2E tests. This RFC and implementation is essentially the wrapper around https://hub.docker.com/r/o1labs/mina-local-network and other tools that is already successfully in use and works really good (including local env. and CI).
Next [o1js](https://github.com/o1-labs/o1js/issues/1169) and [zkapp-cli](https://github.com/o1-labs/zkapp-cli/issues/478) tests will continue utilize same toolset as it is right now (check CI jobs of repos mentioned).
Should we add more capabilities to `zkapp-cli` and allow users interact with deployed zkApp via UI? No problem, since we already have all the tools.

0010-zkapp-cli-lightnet-management.md
shimkiv: I also updated the Docker Hub repo description mentioned above ([link](https://hub.docker.com/r/o1labs/mina-local-network)) in order to provide better services topology overview.

shimkiv:
> Another question I have is related to fault tolerance. How should we plan around node failures? If a node fails, is there any restart policies that happen automatically? If not, could we detect node errors and preemptively fix them? Should we notify users iof these node failures right away? What strategy would you like to see?

Absolutely make sense but I'd extract it as task for later. For now I'd just add `zk lightnet status` simple command to see the network status (running/not running and if not then what was the issue if available).
Today, if something will go wrong with Mina Daemon inside the container then this container will simply stop. I'll check if docker engine has any relevant hooks we can use in order to do some meaningful actions before container will die.

0010-zkapp-cli-lightnet-management.md
```diff
@@ -0,0 +1,106 @@
+# RFC for zkapp-cli enhancements to provide the local lightnet networks management
```
barriebyron, line 1:
```suggestion
# RFC for zkApp CLI enhancements to provide the local lightnet networks management
```

shimkiv:
@bkase can you please approve it?
Implementation differs a bit but not too much and in better way.

bkase (approved):

PR #42, opened 2024-01-16
Vinegar RFC

volhovm:
This introduces the RFC describing vinegar, the compatibility layer for pickles.

PRs merging documentation for pickles created when sketching this RFC:
- https://github.com/o1-labs/proof-systems/pull/1693
- https://github.com/o1-labs/proof-systems/pull/1694


00xx-vinegar.md
```diff
@@ -0,0 +1,101 @@
...
+    - volhovm: what is this "checking that the evaluations satisfy the combined constraint polynomial"? I don't think there's any *checking* or *verification* happening at this point, we merely compute thes polynomial, no?
+1. Verify that the polynomial commitments and their evaluations  are consistent by checking the 'opening proof' of the polynomial commitment scheme.
+    - This corresponds to `OpeningProof::verify` call in the end of `batch_verify`. Internally, this will combine all the evaluations together, and check that they are consistent with the commitments, jointly.
+
+In order to *recursively* verify a kimchi proof in pickles, we need to be able to efficiently run these steps inside a circuit. However, some of the operations are run on elliptic curve points -- represented as tuples in the *base field* -- and others are run against the *scalar field*, that is the field $F_q$ of scalars for elliptic curve points that form a group. To handle both without an expensive simulation of foreign field arithmetic in the verifier circuit, instead we generate a pair of circuits, where each shares some of the work.
```
mrmr1993, line 20: The exposition about the fields here is redundant with the exposition 2 paragraphs earlier. You can delete it from here and simplify this part.

00xx-vinegar.md
```diff
@@ -0,0 +1,101 @@
...
+
+Pickles is our current implementation of a recursive verifier for the kimchi proof system, implemented as a pair of circuits Step and Wrap implemented on top of kimchi.
+
+The Step circuit is a kimchi proof over the Vesta curve (its commitments are Vesta curve points); the Wrap circuit is a kimchi proof over the Pallas curve.
```
mrmr1993, line 24: More relevant here than `its commitments are Vesta curve points` is that `its witness values are Vesta scalar field elements`; fine to mention both though.

00xx-vinegar.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Verifying a proof in a circuit
+
+Assume that we are verifying a Pallas (wrap) proof, but the below applies equivalently to Vesta. Here is how the list in the previous section maps on our curves:
```
mrmr1993, line 28:
```suggestion
Assume that we are verifying a Pallas (wrap) proof, but the below applies equivalently to Vesta (with step and wrap swapped below). Here is how the list in the previous section maps on our curves:
```

00xx-vinegar.md
```diff
@@ -0,0 +1,101 @@
...
+
+1. The logic for absorbing the commitments is run inside a Vesta proof.
+    - Wrap commitments are Pallas points made of two Pallas base field elements. These are Vesta scalar field elements, which makes them easy to reason about in the Vesta proof -- the next Step proof that verifies a Wrap.
```
mrmr1993, line 31: This is a weird place to intro that `vesta_{scalar,base} = pallas_{base,scalar}`. Might be worth extracting to a dedicated paragraph, whose gist is:
* pallas and vesta are a cycle of curves: the scalar field of vesta is the base field of pallas, and the scalar field of pallas is the base field of vesta
* pickles Wrap proofs use the pallas curve, and witness values use the pallas scalar field / vesta base field
* pickles Step proofs use the vesta curve, and witness values use the vesta scalar field / pallas base field.

Then, this can be simplified to
```suggestion
    - The commitments are formed of Vesta scalar field elements, so we can use native arithmetic.
```
or some such.

00xx-vinegar.md
```diff
@@ -0,0 +1,101 @@
...
+  - Also `wrap_verifier/incrementally_verify_proof`, but now again for the VinegarStep. Our VinegarWrap, unlike Wrap, wraps two proofs at a time.
+* Expose the remaining unverified information from 'VinegarStep' in the public input, so that it can be finalized by a pickles step proof
+    - During 'VinegarWrap', the VinegarStep's `challenge_poly_commitments` and `old_bulletproof_challenges` will create `MessageForNextWrap` that will be a part of the VinegarWrap statement. Then they will be passed to the Step proof, and from there consumed in `step_verifier/finalize_other_proof`.
+* Expose a public input compatible with the pickles format, so that pickles recursion can operate over the proof.
+
+The operations needed are conceptually very similar to the operations in the existing pickles circuits. However, because we want to avoid hard-coding the particular details of the target kimchi proof into the circuit, we will need to be careful to construct 'VinegarStep' and 'VinegarWrap' in a generic way, with some mechanism to describe the verification steps required for a particular target kimchi-style proof.
+
+#### Scratch space; incomplete/draft notes
```
mrmr1993, line 86: Delete this section, this is from my original notes when I was drafting.

mrmr1993 (changes requested):
Broadly looks good, apart from a few structural comments. There are a few issues though:
* this doesn't follow the issue template, and essentially only addresses the `Detailed design` section thereof;
* this doesn't have an RFC number (although I can quickly allocate one when this is ready to merge).

volhovm:
@mrmr1993 Thank you for the feedback. 
* Will address the comments and introduce the template -- didn't realise there was one at the point of writing.
* Regarding the number -- yes, I thought we just assign a number at merge.

00xx-vinegar.md
volhovm: Will create a separate paragraph. This was not intended to introduce the concept of curves. I'm just repeating the same thing several times in different phrasings/levels of elaboration which is supposed to help the reader to double-check there's no mistake (because everything is symmetric).

0001-template.md
```diff
@@ -23,7 +23,7 @@ In general:
 
 * Be specific. This document is meant to share intent to your colleagues. Share what you believe you will actually do.
 * Be decisive. No maybes. Any uncertainty can be captured in the unresolved questions section at the end.
-* Provide design contex so that we can align on and commit to a technical design.
```
volhovm, line 26: For the reviewers: changes in `001-template.md` are orthogonal, but very cosmetic, and it would be very hard to merge a separate PR because it needs a lot of approvals. So I want to add it here. It's a separate commit, so can be reverted individually if needed.

PR #43, opened 2024-01-22
Add the RFC for large MSM computation

volhovm:


00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -0,0 +1,238 @@
...
+```
+x = x_0 + 2^k x_1 + 2^(2k) x_2 + ...
+```
+and computing the scaling `x G` in parts as
+```
+x_0 G + x_1 (2^k G) + x_2 (2^(2k) G) + ...
+```
+where each of the `2^i G` terms are known constants.
+
+From here, we can use a standard 'bucketing' trick with `2^k` buckets to avoid doing any doublings or scalings at all, requiring only `n*(254/k)+2^(2k)` curve point additions:
```
mitschabaude, line 112: it's 2 additions per bucket, expect for the zero bucket, so the precise number is $2^{k + 1} - 2$
```suggestion
From here, we can use a standard 'bucketing' trick with `2^k` buckets to avoid doing any doublings or scalings at all, requiring only `n*(254/k)+2^(k + 1)` curve point additions:
```

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -0,0 +1,238 @@
...
+```rust
+let mut buckets: [C; 2^k] = [H; 2^k]; /* Where `H` is the blinding generator. */
+for (coefficient, commitment) in to_scale_pairs {
+    buckets[coefficient] += commitment;
+}
+let mut total = -H.scale(2^k * (2^k + 1) / 2); /* TODO(mrmr1993): Check that this is correct. */
```
mitschabaude, line 118: looks more like `(2^k - 1) * 2^k / 2` to me, since the sum below goes from `1` to `2^k - 1`

00XX-efficient-msms-for-non-native-pickles-verification.md
volhovm: The formula was altogether wrong... The correct number of additions is at most $3n$ since $n$ is the size of MSM and thus the maximum number of buckets.

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -0,0 +1,481 @@
...
+
+### Algorithm Performance, Circuit Layout, and Multi-Circuit Folding
+
+The circuit that we are implementing is large, and its layout should be carefully considered. Luckily, we have some liberty in design decisions -- the algorithm needs to be implemented on a *variant* of Kimchi which can allow long rows, additive lookups (memory access), and support for folding. Two last features are (being) developed in the zkVM project.
+
+Note that each iteration (one run) of the sub-MSM algorithm with limited buckets takes $3n$ additions at most:
+- The cycle in the beginning populates the buckets and it takes $n$ additions because that's how much multiplications are there in `to_scale_pairs`.
+- The end loop does two additions per each iteration (and each iteration assumes a non-zero bucket, assume there are $n'$ of these buckets), so we have $2 \cdot n'$ additions. Since $n' < n$, the worst case is $2 n$ additions.
+
+In addition to $3n$ addition per sub-MSM run, we will need to sum all the results of it. Luckily, this computation is relatively cheap --- we require one extra addition per run.
```
mrmr1993, line 273: My intention was to share the lookup argument across the folds (i.e. delay and run it at the end). Don't we incur a large witness overhead otherwise?

Doing it this way also removes all of those extra additions.

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -0,0 +1,481 @@
...
+
+In total, the whole algorithm then requires $3n \cdot l$ additions because we need to run the inner algorithm $l$ times, and then sum the results (which is negligibly small and can be accumulated along the way). Assuming worst (of the two) case of $n = 2^{16}$ and $k = 15$, we get about $2^{17}$ additions, which is still more than $2^{16}$ budget. But even with the easier $n = 2^{15}$ we still have to fit $3$ times more FF EC additions than there are rows available.
```
mrmr1993, line 275: See above. What we actually want is `l * n + 2*n`. What you're describing is roughly 2.5x worse that the original design.

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -0,0 +1,481 @@
...
+
+
+With that in mind, we will use folding to split the total computation into chunks that fit into our $n$-element SRS. Each folding repetition will externally look like an elliptic curve addition w.r.t. the total accumulator. We will elaborate on the design of a concrete circuit later, but so far we can assume that the only shared states between the rounds of folding are (1) total accumulator for the computed value $\sum\limits_{j=1}^i \mathsf{subres}_j$, (2) RAM lookups. Note that the RAM lookups are not ZK and they don't need to be.
```
mrmr1993, line 278: > Each folding repetition will externally look like an elliptic curve addition w.r.t. the total accumulator.

This also becomes false.

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -0,0 +1,481 @@
...
+
+
+Now, there are several approach that can be taken in terms of a concrete circuit design, and deciding on the approach is part of the implementation effort itself since it is (arguably) too complicated for the RFC. The approaches are as follows:
```
mrmr1993, line 281: > Now, there are several approach that can be taken in terms of a concrete circuit design, and deciding on the approach is part of the implementation effort itself since it is (arguably) too complicated for the RFC.

This is the *whole point* of the RFC. Delete.

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -0,0 +1,481 @@
...
+1. Making the circuit wide enough to fit the whole sub-MSM algorithm in one circuit.
```
mrmr1993, line 282: Not part of the target design; move to alternatives section or delete.

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -0,0 +1,481 @@
...
+
+Regarding ECC, o1js also contains an implementation that is part of ECDSA library. It is unclear though whether this can be useful in the kimchi context, as this library is written in o1js/typescript --- probably it can be used as an inspiration.
+- https://github.com/o1-labs/o1js/blob/main/src/lib/gadgets/elliptic-curve.ts#L99
+
+Additionally, one will need to decide which coordinates would are better to use. Available options are:
+- [Affine](https://www.hyperelliptic.org/EFD/g1p/auto-shortw.html)
+- [Jacobian](https://www.hyperelliptic.org/EFD/g1p/auto-shortw-jacobian.html#doubling-dbl-2007-bl)
+- [Projective](https://www.hyperelliptic.org/EFD/g1p/auto-shortw-projective.html)
+
+Again, **in practice** affine coordinates will be most likely the most efficient, and also easy to implement, so the suggested path is to start by implementing those, and revisiting the technique much later in the project timeline.
```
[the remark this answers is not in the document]
volhovm: Will move to alternatives. It's not less useful than other stuff in the background I think. Good to understand the scope of the issue and rationale.

00XX-efficient-msms-for-non-native-pickles-verification.md
mrmr1993: Your understanding of the protocol was wrong, not the formula :man_facepalming: Please check next time before changing things.

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -0,0 +1,481 @@
...
+- https://hackmd.io/@arielg/B13JoihA8
+
+
+### Algorithm Performance, Circuit Layout, and Multi-Circuit Folding
+
+The circuit that we are implementing is large, and its layout should be carefully considered. Luckily, we have some liberty in design decisions -- the algorithm needs to be implemented on a *variant* of Kimchi which can allow long rows, additive lookups (memory access), and support for folding. Two last features are (being) developed in the zkVM project.
+
+Note that each iteration (one run) of the sub-MSM algorithm with limited buckets takes $3n$ additions at most:
+- The cycle in the beginning populates the buckets and it takes $n$ additions because that's how much multiplications are there in `to_scale_pairs`.
+- The end loop does two additions per each iteration (and each iteration assumes a non-zero bucket, assume there are $n'$ of these buckets), so we have $2 \cdot n'$ additions. Since $n' < n$, the worst case is $2 n$ additions.
```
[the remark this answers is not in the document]
volhovm: True, please ignore this, it was written without the blinding factor in mind. Worst case estimate is the same, but I'll remove the paragraph.

00XX-efficient-msms-for-non-native-pickles-verification.md
volhovm: Fair -- misread the size of the array, will have to update the estimates paragraph.

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -0,0 +1,483 @@
...
+* evaluate the proof-system's constraints at the evaluation point, divided by the 'vanishing polynomial', using the evaluations of each of the columns;
+* absorb the evaluations of each of the columns into the sponge;
+* squeeze a challenge for combining polynomials, and another for combining evaluations;
+* compute the ['combined inner product'](https://github.com/o1-labs/proof-systems/blob/cfc829220b44c1122863eca0db411560b99d6c8e/poly-commitment/src/commitment.rs#L449) of the evaluations.
+
+Verifying the 'opening proof':
+* resume using the 'pre-evaluation sponge', which operates over the base field of `C`;
+* absorb the combined inner product into the sponge;
+* for each of the `log2(domain_size)` rounds, absorb the 'left' and 'right' commitments, and squeeze a challenge for 'folding' the split commitments together;
+* compute the evaluation of $h(X)$ (formed from the folding challenges) at each evaluation point, combined by powers of the evaluation combiner;
```
mrmr1993, line 93: Don't we call `h(X)` `b(X)` in the code? I don't really mind either way though

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -0,0 +1,483 @@
...
+This suggests that we can first accumulate the values of $`\hat G_{c,j}`$ into the correct "buckets", where each bucket corresponds to a coefficient $c_i \in [0,2^k-1]$, and then sum all of them scaling by $c_{i}$.
+
+A lookup table will be used to fetch the corresponding $G_i \cdot 2^{j \cdot k}$. Therefore, the only operations that we need to encoded is the addition of $`\mathbb{F}_{base}(\mathrm{Vesta})`$ elements in $`\mathbb{F}_{scalar}(\mathrm{BN254})`$. Note that the elements $G_i \cdot 2^{j \cdot k}$ will have coordinates in $`\mathbb{F}_{base}(\mathrm{Vesta})`$. Therefore, the table will require more than one limbs for each coordinates.
+
+Assuming the additive notation for the group, the formal code description of the algorithm is as follows:
+
+```rust
+fn fill_buckets(coeffs: Vec<Field>, bases: Vec<Field>, k: uint, H: Group) {
+    for (coefficient, commitment) in coeffs.iter().zip(bases.iter()) {
+        buckets[coefficient] += commitment;
```
mrmr1993, line 172: Not important, but `buckets` isn't defined anywhere

mrmr1993: And `k`, `H` are unused

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -0,0 +1,483 @@
...
+    }
+}
+
+/// Computes the total MSM: \sum_{i=1}^N coeffs_i \cdot bases_i
+fn compute_msm(coeffs: Vec<Field>, bases: Vec<Group>, k: uint, H: Group) {
+    // Initialize the buckets with the blinding factor H
```
mrmr1993, line 178: It might have been my mis-wording, but: `H` is the blinding element, to which the blinding factor is applied.

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -0,0 +1,483 @@
...
+    let mut buckets: [C; 2^k] = [H; 2^k];
+    for i in 0..l=254/k {
```
mrmr1993, line 180: Not important, but this fills an extra limb with `0`s for every term when `k | 254`

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -0,0 +1,483 @@
...
+Finally, to assert that the final computed `buckets[0]` value is what we expect, we need to slightly adjust the additive lookup argument to externally (by modifying lookup boundary conditions) assert that the accumulated computation result (the discrepancy) is contained in the last constraint of the last folding iteration.
+- Our additive lookup constraint has a form of like $\frac{1}{r + v}$, and we can use an alternative accumulator boundary condition $`\mathsf{acc}' = \mathsf{acc} +
+\frac{1}{r + 0} - \frac{1}{r + v_{0}}`$ embedded into the constraint, where $v_0$ is the value contained in the zero slot. In such a way enforce $v_0$ to be present at the zero address. This approach is already taken in the zkVM implementation.
+
+
+
+### Additive Lookups
+
+In the zkVM project there is already an implementation of an additive lookup algorithm.
+- We must make sure, when moving it into our project, that it does not rely heavily on pairings -- even though BN254 supports a pairing, it is better (for future compatibility with IPA Kimchi) to ignore this.
```
mrmr1993, line 305: These wasn't a design considerations. The consideration was that we could 'update the table state' in the style of RAM, which pairing-based lookups make difficult and expensive.

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -0,0 +1,483 @@
...
+- We must make sure that the lookup argument supports lookups of the $k$ bit size, so the parameter must align well with the lookup protocol.
+
+### Folding's IVC
+
+For folding to work correctly "in its full recursive power", additionally to merely instance folding itself, we will need an IVC (interactive verifiable computation) part that will verify the previous folding iteration within the circuit.
+- We consider two approaches regarding folding IVC: either supporting a cycle of curves (BN254 / Grumpkin cycle) or non-native emulation.
```
mrmr1993, line 311: We no longer consider 2 approaches.

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -0,0 +1,483 @@
...
+    - The cycle of curves approach relies on switching between two curves on every folding iteration. Verification of KZG can be encoded with Grumpkin curve (see [this aztec blog post](https://hackmd.io/@aztec-network/ByzgNxBfd#2-Grumpkin---A-curve-on-top-of-BN-254-for-SNARK-efficient-group-operations)).
```
mrmr1993, line 312: Delete

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -0,0 +1,483 @@
...
+    - The non-native emulation uses foreign field arithmetics to proceed. (Note that this is /not/ the Goblin Plonk approach)
+    - We expect to start the project by implementing the FFA/FFEC circuit first. When it is time to start implementing folding with IVC, if only the curves approach is available, it will be our first choice. However the second approach is strongly preferred in the long run.
```
mrmr1993, line 314: Delete

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -0,0 +1,483 @@
...
+    * Outline the general approach or strategy for testing that will be followed. This can include mentioning the types of testing to be performed (e.g., unit testing, integration testing, performance testing) and any specific methodologies or tools that will be utilized.
+3. Testing scope:
+    * Define the scope of testing by identifying the key areas or functionalities that will be covered by testing efforts.
+4. Testing requirements:
+    * Specify any specific testing requirements that need to be considered, such as compliance requirements, security testing, or specific user scenarios to be tested.
+5. Testing resources:
+    * Identify the resources required for testing, such as testing environments, test data, or any additional tools or infrastructure needed for effective testing.-->
+
+Phase 1:
+1. Investigate existing approaches to FFA (foreign field arithmetics) and FFEC (foreign field elliptic curves), including ones implemented for the standard Kimchi. Get basic insight.
```
mrmr1993, line 373: We were never going to do this. Delete.

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -0,0 +1,483 @@
...
+    - Examine the avaliable FFEC implementations / algorithms.
+      - This includes the existing implementation of ECDSA / FFEC in o1js. See if it can be used as a comparison for our implementation -- e.g. as a baseline, if it's performant enough.
+      - But also other implementations (not part of Kimchi/o1js).
+    - Implement the addition of points of (non-native) Vesta with (native) BN254(Fp) using the existing FFA library within Kimchi.
+      - The ECC addition operation must be implemented from scratch, but it relies on the foreign field additions/multiplications support which already exists in Kimchi.
+      - Start with the affine coordinates. If time allows, see if projective or Jacobian can be more performant.
+      - Important: we need to verify that the conditions are respected for the scalar field of BN254. Initially, it has been written for the scalar field of Vesta/Pallas.
+    - Evaluate the complexity of the implementation
+       - Check the number of constraints, proof size, verifier time, etc. Write benchmarks.
+1. Decide on the optimal algorithm for (primarily) FFA, and FFEC.
```
mrmr1993, line 385: As above, so below

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -0,0 +1,483 @@
...
+1. Implement the MSM algorithm in the circuit suggested above. Test and benchmark.
+   - The MSM algorithm for now can be implemented without folding in mind. That is, the circuit can be wider, and maybe it can pass more data through inputs, and verify only parts of the MSM of smaller MSM sizes. The point is to have the algorithm working.
+1. First milestone: Releasing a reduced many-proofs version of MSM algorithm.
+   - This task suggests implementing our MSM algorithm /without/ folding.
+   - Instead, we try to verify "small MSM" where everything fits within one circuit. Then we can release $l$ (or $3l$, e.g. 32 or more) independent proofs that will verify *parts* of the MSM.
+
+
+Phase 2:
+1. Bring folding with IVC into our variant of Kimchi.
+   - If available, use FF IVC (preferred), otherwise the BN254/Grumpkin cycle (temporary fallback).
```
mrmr1993, line 406: Delete

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -0,0 +1,483 @@
...
+   - Analyze folding and try to use it in a simple circuit with one of the Pasta curve. Must be able to prove and verify a circuit. It is independent of this work.
+1. Implement the MSM algorithm with folding and IVC.
+   - The MSM circuit from the previous step should be now properly split into sections and folded.
+1. Evaluate the system and potentially optimise the algorithms.
```
mrmr1993, line 410: This is `Phase 3`, we will not be doing this as part of `Phase 2`

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -0,0 +1,483 @@
...
+
+
+## Drawbacks
+
+1. Cloned Kimchi might diverge with the main variant.
+    - Implementing the whole project in a clone of Kimchi can be problematic to unify with the main version if two diverge significantnly.
+    - Alternatively, one could modify the original Kimchi to support arbitrary number of columns first, support additive lookups, and implement folding. Or, perhaps, at least any of the three. This leads to a more elegant and unified solution and contributes to the main proof system that the company implements.
+    - However this approach becomes a bottleneck for delivering the results, since Kimchi needs to be integrated with Pickles and support backwards compatibility.
+    - The future plan is to generalize existing Kimchi so that we can just *use* it in this project. However at the moment customizability of Kimchi is still WIP, so it is more optimal to start with a clone of Kimchi, and unify them later. Same goes for folding and logups (unless it is easy to reuse them instead of cloning).
```
mrmr1993, line 420: Did you mint a new term in `logups`?

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -0,0 +1,483 @@
...
+1. Current plan does not seem to imply releasing sub-algorithms as independent libraries that might be of value. The whole project is quite complex but at the same highly tailored to one task.
+   - Again, perhaps time constraints are inevitable, and perhaps the code can be later reused and generalised for public use.
+   - The MSM verification task is arguably very important, and even though it's highly tailored to Pickles it might be of big value not just practically, but also as a proof-of-feasibility.
```
mrmr1993, line 423: This doesn't belong in the RFC.

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -0,0 +1,483 @@
...
+
+Prior art is any evidence that your feature (invention, change, proposal) is already known.
+
+Think about the lessons from other blockchain projects or similar updates and provide readers of your RFC with a fuller picture. If there is no prior art, that is fine. Your ideas are interesting whether they are new or adapted from another source.-->
+
+## Unresolved questions
+
+1. Engineering problem: how exactly does the accumulation trick with the additive lookup tables work?
+   - If we go with the circuit layout that fits $2^{15}$ additions into $2^{15}$ rows having exactly one addition per row, we need to make sure the total joint accumulator (between folding iterations) is being updated properly. This requires altering the folding protocol checks.
+2. When splitting MSM into chunks --- is it not problematic that we have to enforce the right order of the chunks? The soundness of the approach relies heavily on the impossibility to swap these chunks around.
```
mrmr1993, line 478: The `buckets_input -> buckets_output` form of the proofs means that there's a canonical order, which is the only one that will work. Suggest deleting, and potentially inlining this note somewhere if important.

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -0,0 +1,483 @@
...
+
+<!--Discuss prior art, both the good and the bad, in relation to this proposal.
+
+Prior art is any evidence that your feature (invention, change, proposal) is already known.
+
+Think about the lessons from other blockchain projects or similar updates and provide readers of your RFC with a fuller picture. If there is no prior art, that is fine. Your ideas are interesting whether they are new or adapted from another source.-->
+
+## Unresolved questions
+
+1. Engineering problem: how exactly does the accumulation trick with the additive lookup tables work?
```
mrmr1993, line 476: Is this still unresolved, given the bullet nested below?

00XX-efficient-msms-for-non-native-pickles-verification.md
volhovm: Used in https://eprint.iacr.org/2023/1284.pdf to describe additive lookups with `1/(beta+f(X))`, so seems somewhat applicable?

00XX-efficient-msms-for-non-native-pickles-verification.md
volhovm: Hm, how do you imagine the drawbacks section looking then? Technical drawbacks only? I can remove the section, agree that these points are a bit contrived, but they were in the template :shrug:

00XX-efficient-msms-for-non-native-pickles-verification.md
volhovm: OK that makes sense, but in the final folded instance, how do we know that we performed exactly `l = 17` buckets iterations, and not less? Unlike in zkVM, we don't have any code pointers...

mrmr1993: We're not folding at this stage.

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -294,24 +295,25 @@ for i in bucket.length/2..buckets.length() {
 ```
 
 Finally, to assert that the final computed `buckets[0]` value is what we expect, we need to slightly adjust the additive lookup argument to externally (by modifying lookup boundary conditions) assert that the accumulated computation result (the discrepancy) is contained in the last constraint of the last folding iteration.
-- Our additive lookup constraint has a form of like $\frac{1}{r + v}$, and we can use an alternative accumulator boundary condition $`\mathsf{acc}' = \mathsf{acc} +
-\frac{1}{r + 0} - \frac{1}{r + v_{0}}`$ embedded into the constraint, where $v_0$ is the value contained in the zero slot. In such a way enforce $v_0$ to be present at the zero address. This approach is already taken in the zkVM implementation.
+- Our additive lookup constraint has a form $`\phi(X) = \phi(X\cdot \omega) + \mathsf{write}(X) - \mathsf{read(X)}`$ where $`\mathsf{read}(\omega^i) = \frac{1}{\alpha + v}`$, and this constraint is run every circuit.
+- While building $`\mathsf{read}`$ and $`\mathsf{write}`$ e can use an alternative accumulator boundary condition $`\mathsf{acc}' = \mathsf{acc} +
```
mrmr1993, line 317: `s/\<e\>/we/`

00XX-efficient-msms-for-non-native-pickles-verification.md
```diff
@@ -473,9 +462,7 @@ Think about the lessons from other blockchain projects or similar updates and pr
 
 ## Unresolved questions
 
-1. Engineering problem: how exactly does the accumulation trick with the additive lookup tables work?
-   - If we go with the circuit layout that fits $2^{15}$ additions into $2^{15}$ rows having exactly one addition per row, we need to make sure the total joint accumulator (between folding iterations) is being updated properly. This requires altering the folding protocol checks.
-2. When splitting MSM into chunks --- is it not problematic that we have to enforce the right order of the chunks? The soundness of the approach relies heavily on the impossibility to swap these chunks around.
+1. When splitting MSM into chunks --- is it not problematic that we have to enforce the right order of the chunks? The soundness of the approach relies heavily on the impossibility to swap these chunks around.
```
mrmr1993, line 487: No.

> The soundness of the approach relies heavily on the impossibility to swap these chunks around.

I don't think this is true/important at the stage we're describing in the RFC. In particular, we have a fixed number `N` of iterations, and for each `0 <= i < N-1` instance `i` has output that's consistent with instance `i+1`'s input.

Suggest deleting.

volhovm: At this stage it's not important, agreed, so will remove :+1: 

Longer-term, when we add folding and IVCs... I would need to understand this more to make sure it's properly sound.

mrmr1993 (approved):

PR #45, opened 2024-01-31
chunking in o1js

Trivo25:


0011-chunking-o1js.md
```diff
@@ -0,0 +1,97 @@
...
+
+In order to make chunking work in o1js, we have to modify the `Pickles.compile` to accept the number of chunks that are required in order to construct the proof. The Pickles [test](https://github.com/MinaProtocol/mina/blob/develop/src/lib/pickles/test/chunked_circuits/test_chunked_circuits.ml) serves as an example on how to modify the `Pickles.compile` function to prove larger circuits and specify the amount of chunks.
+
+Calculating how many chunks are needed to prove a circuit depends on the amount of constraints used in a circuit. Before passing the parameter to the function and compiling the circuit, we have to calculate the amount of chunks needed. Luckily, o1js already has a function that can inspect methods and count the amount of constraints, we get this information by calling `.analyzeMethods()` on the `ZkProgram` or `SmartContract`.
+
+### Technical Implementation
+
+The implementation should not require and API changes for the developer, all changes happen internally and hidden from the public facing API. A naive overview of the changes:
```
nicc, line 27:
```suggestion
The implementation should not require any API changes for the developer, all changes happen internally and are hidden from the public-facing API. A naive overview of the changes:
```

0011-chunking-o1js.md
```diff
@@ -0,0 +1,97 @@
...
+2. Adapting the compile function in [proof_system.ts](https://github.com/o1-labs/o1js/blob/0d82ad1d67417c72b06fc8456a0a2b54bcc46652/src/lib/proof_system.ts)
+
+   - `compileProgram` function should accept a new argument `numChunks` which represents the amount of chunks needed to compile the circuit
+   - `numChunks` inside of `compileProgram` will then be passed into `Pickles.compile` via the bindings layer
+
+3. Before calling into `compileProgram`, the number of chunks must be calculated. This can be done by utilizing the `analyzeMethods` function which exists on both `SmartContract` and `ZkProgram`.
+   - For `ZkProgram`, this change can be done in `proof_systems.ts` and the `ZkProgram` function
+   - For `SmartContract`, the change is needed in [zkapp.ts](https://github.com/o1-labs/o1js/blob/0d82ad1d67417c72b06fc8456a0a2b54bcc46652/src/lib/zkapp.ts) - in the `SmartContract.compile` method.
+
+- ## Changing the type definitions of the bindings to represent the change in the `pickles_compile` function
```
nicc, line 42: Should this be point 4 under the Technical Implementation heading? If so, this has a bit less context than the others. Genuine question: should we point out the types that need to change or is that too much detail?

0011-chunking-o1js.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Test plan and functional requirements
+
+1. Testing Goals and Objectives:
+
+- o1js can compile and prove circuits that are bigger than 2^16 constraints, up to ideally 2^32 (the new limit introduced by chunking)
+- A proof using a larger circuit limit can be deployed to and verified on a testnet
```
nicc, line 49: By 'larger' do you mean > 2^16 and < 2^32? I assume so. Could also be interpreted as > 2^32. Might help to make this explicit

0011-chunking-o1js.md
```diff
@@ -0,0 +1,97 @@
...
+
+- multiple tests in o1js that test the new limits of the circuits. Tests in Kimchi and Pickles already exists, they can be used as [examples](https://github.com/MinaProtocol/mina/blob/develop/src/lib/pickles/test/chunked_circuits/test_chunked_circuits.ml)
+
+3. Testing Scope:
+
+- Proofs can be constructed and verified locally as well as deployed to a testnet.
+
+## Drawbacks
+
+Proving larger circuits will increase proving time and memory consumption. Given the memory limit in most browsers and WASM, this could cause problems for users trying to prove very large circuits.
```
nicc, line 62: @garwalsh fyi

0011-chunking-o1js.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Rationale and alternatives
+
+Not exposing Chunking and continuing to rely on recursion for large computations. This adds a significant overhead for developers and adds complexity of applications.
+
+## Prior art
+
+Chunking working in [OCaml and Pickles](https://github.com/MinaProtocol/mina/blob/develop/src/lib/pickles/test/chunked_circuits/test_chunked_circuits.ml) directly.
+
+## Unresolved questions
```
nicc, line 72: It looks like some additional work may be required in pickles to resolve these issues. If so, we should add these to the Technical Implementation section above to account for all the work. The added implementation items can simply give a brief overview of the neceessary fix and refer to this section (Unresolved questions) for further info

0011-chunking-o1js.md
```diff
@@ -0,0 +1,95 @@
...
+
+In the TypeScript layer, [compiling circuits](https://github.com/o1-labs/o1js/blob/16e66f9fba51ac4832691c3fc7a52bae60a08fd7/src/lib/proof_system.ts#L683-L688) calls into the bindings layer via `Pickles.compile`, passing in the individual branches (rules or methods) of the smart contract or ZkProgram and returning prover artifacts.
+
+In order to make chunking work in o1js, we have to modify the `Pickles.compile` to accept the number of chunks that are required in order to construct the proof. The Pickles [test](https://github.com/MinaProtocol/mina/blob/develop/src/lib/pickles/test/chunked_circuits/test_chunked_circuits.ml) serves as an example on how to modify the `Pickles.compile` function to prove larger circuits and specify the amount of chunks.
+
+Calculating how many chunks are needed to prove a circuit depends on the amount of constraints used in a circuit. Before passing the parameter to the function and compiling the circuit, we have to calculate the amount of chunks needed. Luckily, o1js already has a function that can inspect methods and count the amount of constraints, we get this information by calling `.analyzeMethods()` on the `ZkProgram` or `SmartContract`.
```
mitschabaude, line 23: It would be nice to write out the formula for computing the chunks.

Note that Pickles adds a small amount of overhead circuit which isn't captured by `analyzeMethods()`. I'm not sure if its necessary to include that in the estimate for the number of chunks.

But it _does_ mean that there's an edge case in which the automatic approach fails: If the Pickles overhead slightly pushes the circuit size over the next power of two and makes it require more chunks than estimated.

To handle a failure like that, I think it's necessary to allow users to override `numChunks` through the `compile()` interfaces

0011-chunking-o1js.md
```diff
@@ -0,0 +1,95 @@
...
+
+On the other hand, setting `num_chunks:8`,`override_wrap_domain:N2` and the loop iteration to `2^19` compiles the circuit and yields a valid proof. (see `valid_proof.ml`)
+
+2 What is the maximum allowed number chunks? From my understanding, setting iteration to `2^20` and `num_chunks:16` requires that the wrap domain must be `16`, but it only accepts `N0 | N1 | N2`. Setting the wrap domain to anything `<= N2` yields an error. How do we deal with large number chunks and constraints? `override_wrap_domain` is of type `Pickles_base.Proofs_verified.t` which only accepts `N0 | N1 | N2`.
+See `large_wrap_domain.ml` for this case. `This circuit was compiled for proofs using the wrap domain of size 15, but the actual wrap domain size for the circuit has size 16`
+
+3 Running a small MVP of Chunking in o1js through the bindings yields the following error: `there are not enough random rows to achieve zero-knowledge (expected: 3, got: 2)`. It seems like some values aren't passed correctly. This also needs to be resolved during implementation.
```
mitschabaude, line 95: Might this be related to the hack/workaround present in the ocaml chunking examples, which is not done in `pickles_bindings.ml`?

```ocaml
(* We must now appease the permutation argument gods, to ensure
   that the 7th permuted column has polynomial degree larger
   than 2^16, and thus that its high chunks are non-zero.
   Suckiness of linearization strikes again!
*)
```

0011-chunking-o1js.md
```diff
@@ -0,0 +1,95 @@
...
+```
+
+This can be reproduced by running `dune exec ./invalid_proof.exe` inside of the test folder `src/lib/pickles/test/chunked_circuits`.
+
+Is this behaviors expected? If so, why and how can we prevent the construction of invalid circuits and proofs in o1js? Changing `num_chunks` and `override_wrap_domain` results in an (expected) compilation error - so the chosen values seem to be logical and correct.
+
+On the other hand, setting `num_chunks:8`,`override_wrap_domain:N2` and the loop iteration to `2^19` compiles the circuit and yields a valid proof. (see `valid_proof.ml`)
+
+2 What is the maximum allowed number chunks? From my understanding, setting iteration to `2^20` and `num_chunks:16` requires that the wrap domain must be `16`, but it only accepts `N0 | N1 | N2`. Setting the wrap domain to anything `<= N2` yields an error. How do we deal with large number chunks and constraints? `override_wrap_domain` is of type `Pickles_base.Proofs_verified.t` which only accepts `N0 | N1 | N2`.
+See `large_wrap_domain.ml` for this case. `This circuit was compiled for proofs using the wrap domain of size 15, but the actual wrap domain size for the circuit has size 16`
```
mitschabaude, line 93: Interesting! I wouldn't even have thought that number of chunks influences the wrap domain size.

0011-chunking-o1js.md
Trivo25: It might, although I did try different configurations with different circuits. If this is something that pops up often, we would need a way to better predict when it happens and a way to communicate that to the user if we cant fix it automatically

Trivo25: note: I did write the circuit in o1js directly and passed the values manually to `Pickles.compile`, but good point: I will try and run the example via the bindings directly from OCaml to see if the issue persists

mitschabaude: I meant that maybe we just need to add that workaround to `pickles_bindings.ml`

mitschabaude: I don't understand what it does, but AFAICT it adds a generic gate with a special kind of witness layout which is different from anything that would be added in a normal circuit

Trivo25: yea, but I also wanna make sure that the circuits I test via the bindings are the same as in the OCaml test, so I'll run that same circuit via the OCaml bindings just to check if the problem is fixed by the workaround or if it's something deeper in the bindings

mitschabaude: Ok. Maybe we're missing proper ZK rows handling in the wasm bindings

0011-chunking-o1js.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Prior art
+
+Chunking working in [OCaml and Pickles](https://github.com/MinaProtocol/mina/blob/develop/src/lib/pickles/test/chunked_circuits/test_chunked_circuits.ml) directly.
+
+## Unresolved questions
+
+While trying to prototype Chunking in o1js, I came up with the following unresolved questions. The tests can be found in [this](https://github.com/MinaProtocol/mina/tree/chunking_rfc_unresolved/src/lib/pickles/test/chunked_circuits) branch.
+
+1 Changing the loop that constructs the circuit from `2^17` (roughly `2^16` constraints) iterations to `2^18` iterations (or `2^17` constraints) and also adjusting `num_chunks: 4` and `override_wrap_domain:N1` compiles the program correctly, generates a proof _but_ doesn't verify the proof:
```
mrmr1993, line 74: Wrap circuit also needs chunking

0011-chunking-o1js.md
nicc: From Matthew:
> The wrap domain size is capped -- and we don't do chunking there -- so there's a limit on the number of chunks we can use until / unless we also chunk the wrap proof

Issue #44, opened 2024-01-30
Fungible Token Standard


garwalsh:
Will build on: https://forums.minaprotocol.com/t/draft-fungible-token-standard-zkapps/6142

In light of the work to generalise the base token here:
https://app.zenhub.com/workspaces/sdk-6130fedb3b0fc600123d8796/issues/gh/o1-labs/o1js/1274

mitschabaude:
PR that implements and describes new `TokenContract` API: https://github.com/o1-labs/o1js/pull/1384

IMO the `Approvable` and `Transferable` APIs, proposed for the token standard [here](https://forums.minaprotocol.com/t/draft-fungible-token-standard-zkapps/6142) should be changes to exactly match the new `TokenContract` API

Here I proposed that we remove the Callback API, which is done by now: https://github.com/o1-labs/o1js/pull/1402
So the token standard should also no longer rely on that API, but on what the TokenContract does: accepting an `AccountUpdateForest` which contains all the account updates that it has to approve, in one structure

I don't have strong opinions on some of the "viewable" APIs in the standard

I don't thing the `Pausable` API proposed there is needed.

Trivo25:
see https://app.zenhub.com/workspaces/sdk-6130fedb3b0fc600123d8796/issues/zh/102

PR #46, opened 2024-03-13
[0014] Fungible Token Standard

kantp:
This introduces an rfc for a fungible token standard, as per #44.

### DIFF
--- a/0014-fungible-token-standard.md
+++ b/0014-fungible-token-standard.md
@@ -18,7 +18,7 @@ Fungible tokens on Mina will use the custom token feature of Mina defined in [MI
 
 The token owner contract can change balances of accounts with the custom token, so it can mint, burn, and move tokens. When an `AccountUpdate` that has not been created in a method of the owner contract tries to modify the balance of an account with the custom token, it needs to be approved by the owner contract. That way, the owner contract can enforce rules that all transactions with the token must satisfy (conservation of tokens being a very common example). The [`TokenContract` class](https://github.com/o1-labs/o1js/pull/1384) defines methods for token transfer, as well as for approving a whole forest of account updates. The API that we define mirrors those methods, and the reference implementation is built using the `TokenContract`.
 
-The API consists of the following `TypeScript` `interface`s:
+The API consists of the following TypeScript interfaces:
 
 ### Transferable
 This interface consists of a single function, `transfer`, which sends a specified amount of tokens from one account to another.
@@ -37,7 +37,6 @@ Transactions that are either more complicated than a transfer from one account t
 
 ```TypeScript
 interface Approvable {
-
   approveAccountUpdate(accountUpdate: AccountUpdate | AccountUpdateTree): void;
   approveAccountUpdates(accountUpdates: (AccountUpdate | AccountUpdateTree)[]): void;
   approveBase(forest: AccountUpdateForest): void;
--- a/0014-fungible-token-standard.md
+++ b/0014-fungible-token-standard.md
@@ -44,7 +44,7 @@ interface Approvable {
 ```
 
 ### Administrative Interfaces
-For administrative actions, such as changing the supply or updating the `VerificationKey`, we define the `Mintable`, `Burnable`, and `Upgradeable` interfaces:
+For administrative actions, such as changing the total supply or minting and burning tokens, we define the `Mintable` and `Burnable` interfaces:
 
 ```TypeScript
 interface Mintable {
@@ -56,10 +56,6 @@ interface Mintable {
 
 interface Burnable {
   burn: (from: PublicKey, amount: UInt64) => AccountUpdate;
-}
-
-interface Upgradable {
-  setVerificationKey: (verificationKey: VerificationKey) => void;
 }
 ```
 
--- a/0014-fungible-token-standard.md
+++ b/0014-fungible-token-standard.md
@@ -28,7 +28,7 @@ This interface consists of a single function, `transfer`, which sends a specifie
 interface Transferable {
   transfer(from: PublicKey | AccountUpdate,
            to: PublicKey | AccountUpdate,
-           amount: UInt64 | number | bigint): void;
+           amount: UInt64): void;
 }
 ```
 
--- a/0014-fungible-token-standard.md
+++ b/0014-fungible-token-standard.md
@@ -84,6 +84,8 @@ We shall extend the test suite to also cover cases that must fail (transactions 
 
 ## Drawbacks
 
+Currently, it is only possible to pass token permissions for one kind of token down an `AccountUpdate` tree. That limitation makes it somewhat cumbersome to implement solutions involving multiple custom tokens (such as a decentralized exchange). Removing that limitation in a future upgrade to `o1js` and Mina would make development around custom tokens easier.
+
 ## Prior art
 
 This standard has evolved from a discussion and proposal on MinaResearch, see https://forums.minaprotocol.com/t/draft-fungible-token-standard-zkapps/6142
--- a/0014-fungible-token-standard.md
+++ b/0014-fungible-token-standard.md
@@ -1,0 +1,98 @@
+# Fungible Token Standard on Mina
+
+## Summary
+
+Fungible tokens are a pervasive feature that are used for many applications, such as bridges, decentralised exchanges, layer 2 solutions, and others. Establishing a standard for creating fungible tokens on Mina will reduce the efforts for builders to create their own tokens, and allow exchanges, explorers, and other third party applications to integrate fungible tokens in a uniform way.
+
+## Motivation
+
+Fungible tokens are a common feature on programmable blockchain systems. Establishing a standard will help developers to create their own fungible tokens, and to interface their application with tokens that others have created.
+
+We propose a modular standard, consisting of a number of `TypeScript` interfaces. These define an API which third parties can target for integration.  We also provide a reference implementation, based on the [custom token feature](https://github.com/MinaProtocol/MIPs/blob/main/MIPS/mip-zkapps.md#custom-tokens) of Mina and the [`TokenContract` class](https://github.com/o1-labs/o1js/pull/1384) of `o1js.`
+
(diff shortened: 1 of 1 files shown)


0014-fungible-token-standard.md
```diff
@@ -0,0 +1,98 @@
...
+  getBalanceOf: (address: PublicKey, options: ViewableOptions) => UInt64;
+  getTotalSupply: (options: ViewableOptions) => UInt64;
+  getCirculatingSupply: (options: ViewableOptions) => UInt64;
+  getDecimals: () => UInt64;
+}
+```
+
+### Reference Implementation
+
+A reference implementation of a fungible token implemeting the above interfaces can be found at https://github.com/MinaFoundation/mip-token-standard.
```
stevenpack, line 82: Where is the code? Index.ts is empty at that link

stevenpack: Ah, this is TBD.

0014-fungible-token-standard.md
```diff
@@ -0,0 +1,98 @@
...
+
+The token owner contract can change balances of accounts with the custom token, so it can mint, burn, and move tokens. When an `AccountUpdate` that has not been created in a method of the owner contract tries to modify the balance of an account with the custom token, it needs to be approved by the owner contract. That way, the owner contract can enforce rules that all transactions with the token should satisfy (conservation of tokens being a very common example). The [`TokenContract` class](https://github.com/o1-labs/o1js/pull/1384) defines methods for token transfer, as well as for approving a whole forest of account updates. The API that we define mirrors those methods, and the reference implementation is built using the `TokenContract`.
+
+The API consists of the following `TypeScript` `interface`s:
```
stevenpack, line 21: I'll let others closer to the code comment also, but these interfaces look pretty clean and understandable to met @kantp 👍

0014-fungible-token-standard.md
```diff
@@ -0,0 +1,98 @@
...
+
+### Transferable
+This interface consists of a single function, `transfer`, which sends a specified amount of tokens from one account to another.
+
+```TypeScript
+
+interface Transferable {
+  transfer(from: PublicKey | AccountUpdate,
+           to: PublicKey | AccountUpdate,
+           amount: UInt64 | number | bigint): void;
```
ymekuria, line 31: The general token contract has the amount type:
`amount: UInt64` 

https://github.com/o1-labs/o1js/blob/b5ef43221392692036e91c864ea64719ff605860/src/lib/mina/token/token-contract.ts#L121-L124

Will this function take the amount imput type of `amount: UInt64 | number | bigint` and convert it to UInt64?

0014-fungible-token-standard.md
```diff
@@ -0,0 +1,98 @@
...
+}
+```
+
+### Approvable
+Transactions that are either more complicated than a transfer from one account to another (such as sending tokens from one account to many accounts) or that are constructed from third-party contracts can be done using the `Approvable` interface. The workflow is to construct the individual `AccountUpdate` values, and then have them be approved by the owner contract. The interface defines three functions to cover the cases of individual `AccountUpdate`s or `AccountUpdateTree`s, an array of either of these types, or a whole `AccountUpdateForest`.
```
ymekuria, line 36: It would be beneficial to briefly demonstrate how more complicated transactions such as sending tokens from one account to many accounts or transactions that are constructed from third-party contracts can be done using this Approvable interface.

ymekuria (commented):
Nice work @kantp! The interfaces in this RFC are generally intuitive and follow the general token contract well. I left some comments, questions, and suggestions.

0014-fungible-token-standard.md
```diff
@@ -0,0 +1,98 @@
...
+
+The token owner contract can change balances of accounts with the custom token, so it can mint, burn, and move tokens. When an `AccountUpdate` that has not been created in a method of the owner contract tries to modify the balance of an account with the custom token, it needs to be approved by the owner contract. That way, the owner contract can enforce rules that all transactions with the token must satisfy (conservation of tokens being a very common example). The [`TokenContract` class](https://github.com/o1-labs/o1js/pull/1384) defines methods for token transfer, as well as for approving a whole forest of account updates. The API that we define mirrors those methods, and the reference implementation is built using the `TokenContract`.
+
+The API consists of the following `TypeScript` `interface`s:
```
Trivo25, line 21: No need to put TypeScript and interfaces into a code block

0014-fungible-token-standard.md
```diff
@@ -0,0 +1,98 @@
...
+           amount: UInt64 | number | bigint): void;
+}
+```
+
+### Approvable
+Transactions that are either more complicated than a transfer from one account to another (such as sending tokens from one account to many accounts) or that are constructed from third-party contracts can be done using the `Approvable` interface. The workflow is to construct the individual `AccountUpdate` values, and then have them be approved by the owner contract. The interface defines three functions to cover the cases of individual `AccountUpdate`s or `AccountUpdateTree`s, an array of either of these types, or a whole `AccountUpdateForest`.
+
+```TypeScript
+interface Approvable {
+
```
Trivo25, line 40: empty space

0014-fungible-token-standard.md
```diff
@@ -0,0 +1,98 @@
...
+  mint: (to: PublicKey, amount: UInt64) => AccountUpdate;
+  setTotalSupply: (amount: UInt64) => void;
+}
+
+interface Burnable {
+  burn: (from: PublicKey, amount: UInt64) => AccountUpdate;
+}
+
+interface Upgradable {
+  setVerificationKey: (verificationKey: VerificationKey) => void;
```
Trivo25, line 63: Probably worth keeping the recent verification key upgrade changes in mind, if you havent already https://github.com/MinaProtocol/mina/blob/develop/rfcs/0052-verification-key-permissions.md#detailed-design

0014-fungible-token-standard.md
```diff
@@ -0,0 +1,98 @@
...
+### Viewable
+For retrieving information about the token, such as the total or circulating supply, account balances, etc., we define the `Viewable` interface.
+
+```TypeScript
+interface Viewable {
+  getAccountOf: (address: PublicKey) => ReturnType<typeof Account>;
+  getBalanceOf: (address: PublicKey, options: ViewableOptions) => UInt64;
+  getTotalSupply: (options: ViewableOptions) => UInt64;
+  getCirculatingSupply: (options: ViewableOptions) => UInt64;
+  getDecimals: () => UInt64;
```
Trivo25, line 76: I suppose decimals will not be a on-chain state and a hard coded constant, correct?

kantp: Right, that will be static for any token, so no need to have it as state.

0014-fungible-token-standard.md
```diff
@@ -0,0 +1,98 @@
...
+
+The token owner contract can change balances of accounts with the custom token, so it can mint, burn, and move tokens. When an `AccountUpdate` that has not been created in a method of the owner contract tries to modify the balance of an account with the custom token, it needs to be approved by the owner contract. That way, the owner contract can enforce rules that all transactions with the token should satisfy (conservation of tokens being a very common example). The [`TokenContract` class](https://github.com/o1-labs/o1js/pull/1384) defines methods for token transfer, as well as for approving a whole forest of account updates. The API that we define mirrors those methods, and the reference implementation is built using the `TokenContract`.
+
```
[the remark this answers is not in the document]
kantp: Right, good point. There was a pointer to a slack discussion on the ticket, can you maybe add me to that channel so I can see it?

0014-fungible-token-standard.md
```diff
@@ -0,0 +1,98 @@
...
+}
+```
+
+### Viewable
+For retrieving information about the token, such as the total or circulating supply, account balances, etc., we define the `Viewable` interface.
+
+```TypeScript
+interface Viewable {
+  getAccountOf: (address: PublicKey) => ReturnType<typeof Account>;
+  getBalanceOf: (address: PublicKey, options: ViewableOptions) => UInt64;
```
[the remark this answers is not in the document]
kantp: Ah, good catch! In the original design, there was an option to skip some `requireEquals` checks. I think we can drop that from the standard and make it a bit simpler.

0014-fungible-token-standard.md
kantp: Thanks! That interface was older than the token contract. I'll change that to use the same type here.

0014-fungible-token-standard.md
```diff
@@ -0,0 +1,98 @@
...
+  setVerificationKey: (verificationKey: VerificationKey) => void;
+}
+```
+
+### Viewable
+For retrieving information about the token, such as the total or circulating supply, account balances, etc., we define the `Viewable` interface.
+
+```TypeScript
+interface Viewable {
+  getAccountOf: (address: PublicKey) => ReturnType<typeof Account>;
```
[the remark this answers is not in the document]
kantp: Good question! I think it would be more useful when provable, would you agree?

0014-fungible-token-standard.md
kantp: Thanks for the pointer, I'll make sure it's compatible with that rfc.

kantp: @Trivo25 If I understand correctly, then upgrading the verification key is not really something that is special for token contracts, it's more general. Would you agree it's best to remove it from this standard?

kantp:
Thanks for all the suggestions! I've made some changes, and I think we're good for a re-review :)

mrmr1993 (commented):
Is there any intention to provide API access for other zkApps to interact with these tokens? As currently designed, a calling zkApp will have no signal that any of the actions in the API actually happened.
In general, I would describe this more as a `token user API standard`, given that it doesn't handle anything about interactions on the Mina zkApp platform.

mitschabaude:
> Is there any intention to provide API access for other zkApps to interact with these tokens?

yeah that was an explicit goal, see

> In particular, this design allows interoperability in the sense that arbitrary third party SmartContracts can use fungible tokens.

mrmr1993 (approved):
Re-reviewed and got some extra info on the o1js side. Seems like this is now all good to go, sorry for the delay @kantp!

PR #50, opened 2024-06-05
Update rfc 14 (fungible tokens)

kantp:
While working on the implementation for fungible tokens, we realised that the design had to change significantly from the rfc in order to be practical. Most importantly, the original plan of having custom contract code for each token would have placed a significant burden on third parties to integrate.

This PR updates the rfc to match the design we ended up with.

### DIFF
--- a/0014-fungible-token-standard.md
+++ b/0014-fungible-token-standard.md
@@ -77,7 +77,7 @@ In order to allow flexibility in granting permisssions, the methods in this sect
 
 #### The Admin Contract
 ```TypeScript
-export type FungibleTokenAdminBase = SmartContract & {
+type FungibleTokenAdminBase = SmartContract & {
   canMint(accountUpdate: AccountUpdate): Promise<Bool>
   canChangeAdmin(admin: PublicKey): Promise<Bool>
   canPause(): Promise<Bool>
@@ -135,17 +135,17 @@ Those methods call `canPause()` and `canResume()` of the admin contract, respect
 Minting, burning, or transferring tokens (via the `transfer()` method, not via approving account updates) will emit the following events:
 
 ```TypeScript
-export class MintEvent extends Struct({
+class MintEvent extends Struct({
   recipient: PublicKey,
   amount: UInt64,
 }) {}
 
-export class BurnEvent extends Struct({
+class BurnEvent extends Struct({
   from: PublicKey,
   amount: UInt64,
 }) {}
 
-export class TransferEvent extends Struct({
+class TransferEvent extends Struct({
   from: PublicKey,
   to: PublicKey,
   amount: UInt64,
--- a/0014-fungible-token-standard.md
+++ b/0014-fungible-token-standard.md
@@ -14,77 +14,135 @@ In order to allow for some flexibility without changing the token contract itsel
 
 We use Mina's [custom token feature](https://github.com/MinaProtocol/MIPs/blob/main/MIPS/mip-zkapps.md#custom-tokens) to store account balances directly in the Mina ledger, as opposed to storing balances in the state of the token contract itself. That way, we do not rely on any particular off-chain storage solution. Furthermore, this helps us to allow for multiple transactions involving a particular token within the same block, which would be more difficult to achieve if balances were part of the contract state.
 
-The standard implementation supports the common features associated with fungible tokens: minting and burning, transferring, and viewing account balances. One-to-one transfers can be initiated by calling a `@method` of the contract defining the token (the _token owner contract_). More complicated transactions can be formed by constructing multiple `AccountUpdate`s and have those be _approved_ by the token owner contract. In particular, this design allows interoperability in the sense that arbitrary third party `SmartContract`s can use fungible tokens.
+The standard implementation supports the common features associated with fungible tokens: minting and burning, transferring, and viewing account balances. One-to-one transfers can be initiated by calling a `@method` of the contract defining the token (the _token owner contract_). More complicated transactions can be formed by constructing multiple `AccountUpdate`s and have those be _approved_ by the token owner contract. In particular, this design allows interoperability in the sense that arbitrary third party `SmartContract`s can use fungible tokens, by integrating against the one standard implementation.
 
 ## Detailed design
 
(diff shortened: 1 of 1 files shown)


0014-fungible-token-standard.md
```diff
@@ -2,85 +2,178 @@
...
+
+#### Query Balances
 ```TypeScript
-interface Mintable {
-  totalSupply: State<UInt64>;
-  circulatingSupply: State<UInt64>;
-  mint: (to: PublicKey, amount: UInt64) => AccountUpdate;
-  setTotalSupply: (amount: UInt64) => void;
-}
+@method.returns(UInt64) async getBalanceOf(address: PublicKey)
```
Trivo25, line 67: adding `: UInt64` to the function signature here would be nice as well

0014-fungible-token-standard.md
```diff
@@ -2,85 +2,178 @@
...
+
+In this section, we list methods that require some sort of special privileges, like minting new tokens. The rules for when it is permissible to mint new tokens will be different for different tokens. A simple rule could be to require a signature from one or more special keys. There could also be a total limit on the number of tokens in existence, or it could be forbidden to mint new tokens at all.
 
-interface Burnable {
-  burn: (from: PublicKey, amount: UInt64) => AccountUpdate;
+In order to allow flexibility in granting permisssions, the methods in this section will call to a token admin contract, which is set during deployment of the token contract. That admin contract can grant or deny the right to mint tokens, pause/resume transfers, or change the admin contract itself. By using a third contract, the permissions can be changed without changing the token contract itself -- which is important for integration.
+
+#### The Admin Contract
+```TypeScript
+export type FungibleTokenAdminBase = SmartContract & {
```
Trivo25, line 80: do we need the `export` here?

0014-fungible-token-standard.md
```diff
@@ -2,85 +2,178 @@
...
 
+#### Deploying the Contract
 ```TypeScript
-interface Viewable {
-  getAccountOf: (address: PublicKey) => ReturnType<typeof Account>;
-  getBalanceOf: (address: PublicKey, options: ViewableOptions) => UInt64;
-  getTotalSupply: (options: ViewableOptions) => UInt64;
-  getCirculatingSupply: (options: ViewableOptions) => UInt64;
-  getDecimals: () => UInt64;
+export interface FungibleTokenDeployProps extends Exclude<DeployArgs, undefined> {
```
Trivo25, line 94: `export` needed here as well?

0014-fungible-token-standard.md
```diff
@@ -2,85 +2,178 @@
...
+  /** Address of the contract controlling permissions for administrative actions */
+  admin: PublicKey
+  /** The token symbol. */
+  symbol: string
+  /** A source code reference, which is placed within the `zkappUri` of the contract account. */
```
Trivo25, line 99: Do we recommend developers where this url should point to? github?

0014-fungible-token-standard.md
```diff
@@ -2,85 +2,178 @@
...
+  src: string
+  /** Number of decimals in a unit */
+  decimals: UInt8
 }
+
+async deploy(props: FungibleTokenDeployProps)
```
Trivo25, line 105: does this belong to the interface? seems a little out of place

0014-fungible-token-standard.md
```diff
@@ -2,85 +2,178 @@
...
+Deploys the token contract. The token admin contract is assumed to be already deployed, at the address given by `admin`. The token symbol, number of digits, and reference to the source code are to be provided.
+
+#### Changing the Admin Contract
+```TypeScript
+@method async setAdmin(admin: PublicKey)
+```
+
+#### Minting Tokens
+```TypeScript
+@method.returns(AccountUpdate) async mint(recipient: PublicKey, amount: UInt64)
```
Trivo25, line 117: lets specify the return type here as well

0014-fungible-token-standard.md
```diff
@@ -2,85 +2,178 @@
...
+After `pause()` has been successfully called, users will not be able to move or burn tokens, until `resume()` has been called successfully.
+
+Those methods call `canPause()` and `canResume()` of the admin contract, respectively, and only succeed on `Bool(true)`.
+
+### Events
+
+Minting, burning, or transferring tokens (via the `transfer()` method, not via approving account updates) will emit the following events:
+
+```TypeScript
+export class MintEvent extends Struct({
```
Trivo25, line 139: export needed?

0014-fungible-token-standard.md
```diff
@@ -2,85 +2,178 @@
...
+```
+
+### Actions and Reducers
+
+The token contract keeps track of the current circulation of tokens. As it is not currently possible to have multiple transactions in one block that modify the same part of the contract state, we use actions and reducers instead.
+
+Minting and burning tokens will dispatch an action to modify the circulating supply. These actions can be reduced by calling
+
+```TypeScript
+@method.returns(UInt64) async getCirculating()
```
Trivo25, line 163: return type

Trivo25 (approved):
Thank you so much for taking the time to 1) adjust the old RFC and 2) figure out a way to build a good token standard that even I am happy with! :D I've just added some minor nits

0014-fungible-token-standard.md
kantp: I think we do, in order to allow users to write their own admin contract that `implements` that type.

0014-fungible-token-standard.md
kantp: Right, this is not necessary, I think.

0014-fungible-token-standard.md
kantp: Github makes sense, I'll add a sentence.

0014-fungible-token-standard.md
kantp: You're right. This was an error while collecting the code snippets from the implementation. Good catch!

0014-fungible-token-standard.md
kantp: Isn't it useful to have them exported in order to allow third parties to analyse the events?

0014-fungible-token-standard.md
Trivo25: But in the code snippet in the RFC? Not a big deal if we keep it btw, I just noticed it

kantp: Ah, the code snippet is copied over from the implementation, so it has all the export statements that the code has.

And you're right, that is more information than you'd want in the rfc, so let me remove the exports from this document.

nholland94 (approved):

PR #51, opened 2024-06-11
RFC for account deletion

rbonichon:
This document describes the account deletion feature, its motivation and proposed solution as well as possible culprits.

### DIFF
--- a/00xx-account-deletion.md
+++ b/00xx-account-deletion.md
@@ -225,6 +225,13 @@ with respect to the `permissions` filed of `Account_update.Update.t`.
 
 ### Updating the archive node
 
+The addition of a field in `Account_update` will need to reflected in table
+`zkapp_account_update_body` with the addition of a `delete_account` column.
+
+The migration of the current database needs to set a `false` value in this new
+column for all already registered rows.
+
+
 
 ### Other expected code changes
 #### Ledgers
--- a/00xx-account-deletion.md
+++ b/00xx-account-deletion.md
@@ -143,14 +143,36 @@ Deletion support relies on the same techniques as in-memory ledgers, tailored to
  !-- mask ledger https://github.com/MinaProtocol/mina/blob/develop/src/lib/merkle_mask/masking_merkle_tree.ml#L973
  !-- db ledger https://github.com/MinaProtocol/mina/blob/develop/src/lib/merkle_ledger/database.ml#L559 -->
 
+
 ## API
 
 ### Ledgers
 
-Removing elements in ledgers is supported by 2 functions :
+We propose to support removing elements in ledgers through  2 functions :
 
 - `val remove_location: t -> location -> unit`
 - `val remove_account: t -> account -> unit`
+
+
+The current ledger interface also exposes the following function
+```ocaml
+(** for account locations in the ledger, the last (rightmost) filled
+    location
+*)
+val last_filled : t -> Location.t option
+```
+
+We propose to change the implemented interface with a pair of functions
+- `last_filled` returns the last filled location (which might not be the righmost one anymore)
+- `rightmost_filled`, a new function, would return the rightmost location that is filled
+
+Alternatively, depending on the current usage of `last_filled`, we could
+implement a different function `next_fillable` that would return the next
+fillable location with the same exact signature.
+
+
+
+
 
 ### o1js
 The API used by `o1js` is not expected to change much but for some details.
--- a/00xx-account-deletion.md
+++ b/00xx-account-deletion.md
@@ -21,7 +21,6 @@ The support for account deletion relies on adding features at the levels of
 transaction logic(s) and storage (ledgers). The latter is a prerequisite of the
 former.
 
-
 ### Storage
 
 Within the protocol-related codebase, the lowest-level support lies in having a
@@ -151,8 +150,7 @@ Deletion support relies on the same techniques as in-memory ledgers, tailored to
 Removing elements in ledgers is supported by 2 functions :
 
 - `val remove_location: t -> location -> unit`
-- `val remove_account: t -> account -> unit
-`
+- `val remove_account: t -> account -> unit`
 
 ### o1js
 The API used by `o1js` is not expected to change much but for some details.
@@ -170,6 +168,25 @@ The API used by `o1js` is not expected to change much but for some details.
   ```
 
 - *recipient* :
+
+
+
+## Test plan
+
+- **Testing goals and objectives**:
+  - Check the correctness of the account deletion feature;
+  - Validate the absence of impact of the feature on the existing functionalities.
+- **Testing approach**:
+  - Unit tests for the ledger implementations;
+  - Integration tests for the extension of transaction logics.
+- **Testing scope**:
+  - Testing sequences of intertwined addition/removal commands;
+  - Testing simple high-level account updates;
+  - Testing from `o1js` typical use-cases scenarios.
+- **Testing requirements**:
+  - Test in various local setups to simulate different development environments.
+- **Testing resources**:
+  - Simulated environments resembling typical local development setups.
 
 
 # Resources
--- a/00xx-account-deletion.md
+++ b/00xx-account-deletion.md
@@ -193,7 +193,7 @@ deletion.
 
 Clearly, account deletion should not be permitted except for authorized keys, e.g., the
 contract that create the consumable account. For that, we need to check this
-with respect to the `permissions` filed of `Account_update.Update.t`.
+with respect to the `permissions` field of `Account_update.Update.t`.
 
 <!-- Locations to track
  !--
@@ -261,8 +261,8 @@ change is relatively massive so that adding the extra field should really be a
 single task to help reviewers.
 
 
-
 #### o1js
+
 The API used by `o1js` is not expected to change much but for some details.
 
 - *accounts update* : we propose to implement account deletion as a specific
--- a/00xx-account-deletion.md
+++ b/00xx-account-deletion.md
@@ -291,6 +291,17 @@ needs to be handle this correctly and this in turn relies on
 The change here will mainly be to call ``set_all_accounts_rooted_at_exn` as many
 times as there are chunks in the data.
 
+#### JSON encoding of ledgers
+
+The current JSON encoding of ledgers does not mention locations. It is a
+straightforward array of accounts. We now need to state where empty accounts are.
+
+We propose to use a similar strategy as above, that is we still send an
+array. However its elements are all elements within the fill frontier, that is:
+- either an account (aka a JSON object); or
+- an integer, which represent the size of a contiguous interval of empty accounts.
+
+For example, the ledger  `[a, X, X, d, X, f, X, X]` would be encoded as `[a, 2, d, 1, f]`.
 
 
 ### Transaction logic and snark
--- a/00xx-account-deletion.md
+++ b/00xx-account-deletion.md
@@ -229,24 +229,26 @@ type definitions depending on `Account_update`, including
 
 Deleting an account $a₀$ results in two actions:
 1. Actual removal of account $a₀$ of the storage layer;
-2. Return of MINA balance from $a₉$ and initial creation fee to a specified $a₁$ account.
+2. Return of balance of MINA tokens from $a₉$ and initial creation fee to a specified $a₁$ account.
 
-At the snark level, the deletion should thus be equivalent to proving two things
+Handling other tokens is the responsability of the smart contract(s) dealing with these tokens.
+
+Furthermore, account deletion should not be permitted except for some authorized
+participants, e.g., the contract that created the consumable account.
+
+We propose to restrict this capacity to the account to be deleted, so that only
+yourself can call for your own deletion.
+
+For that, we need to check this with respect to the `permissions` field of
+`Account_update.Update.t`.
+
+
+At the snark level, deletion should thus be equivalent to proving two things
 1. The location $l$ of account $a₀$ is now  equal to the empty account.
-   This should be the same proof as calling `set_account` to location $l$ with the empty account;
+   A simple implementation is to call [`set_account`](https://github.com/MinaProtocol/mina/blob/4495af5caea5e1bb2f98f92592c065f93a586ade/src/lib/transaction_snark/transaction_snark.ml#L1531) to location $l$ with the empty account;
 2. The change in MINA of $a₁$ should be equal to
    - the initial creation fee;
    - augmented by any remaining MINA balance.
-
-Furthermore, account deletion should not be permitted except for some authorized
-participants, e.g., the contract that created the consumable account. We propose
-to restrict this capacity to the account to be deleted, so that only yourself
-can call for your own deletion.
-
-For that, we need to check this
-with respect to the `permissions` field of `Account_update.Update.t`.
-
-Now similar to account creationg, one needs to check the resulting balance.
 
 
 <!-- user command snark https://github.com/MinaProtocol/mina/blob/develop/src/lib/transaction_logic/mina_transaction_logic.ml#L1007
--- a/00xx-account-deletion.md
+++ b/00xx-account-deletion.md
@@ -160,30 +160,10 @@ Concretely, the basic support is provided by adding a deletion flag on account
 updates, i.e., extend the `Account_update.Body.Stable.t` type with an additional
 record field `delete_account`.
 
-This turns the actual definition from
-```ocaml
-      type t = Mina_wire_types.Mina_base.Account_update.Body.V1.t =
-        { public_key : Public_key.Compressed.Stable.V1.t
-        ; token_id : Token_id.Stable.V2.t
-        ; update : Update.Stable.V1.t
-        ; balance_change :
-            (Amount.Stable.V1.t, Sgn.Stable.V1.t) Signed_poly.Stable.V1.t
-        ; increment_nonce : bool
-        ; events : Events'.Stable.V1.t
-        ; actions : Events'.Stable.V1.t
-        ; call_data : Pickles.Backend.Tick.Field.Stable.V1.t
-        ; preconditions : Preconditions.Stable.V1.t
-        ; use_full_commitment : bool
-        ; implicit_account_creation_fee : bool
-        ; may_use_token : May_use_token.Stable.V1.t
-        ; authorization_kind : Authorization_kind.Stable.V1.t
-        }
-```
-
-into
-
+This turns the current definition into
     ```ocaml
-      type t = Mina_wire_types.Mina_base.Account_update.Body.V1.t =
+      (* This will update the version of the correspond Mina_wire_types definition too).
+      type t = Mina_wire_types.Mina_base.Account_update.Body.V2.t =
         { public_key : Public_key.Compressed.Stable.V1.t
         ; token_id : Token_id.Stable.V2.t
         ; update : Update.Stable.V1.t
--- a/00xx-account-deletion.md
+++ b/00xx-account-deletion.md
@@ -441,7 +441,7 @@ specify below, using types definde in the ledger interface
 
 - `val remove_location: t -> Location.t -> unit`: remove the account found at the
   given `location` from the ledger, don't do anything if `location` is occupied
-  by the empty account
+  by the empty account;
 - `val remove_account: t -> account_id -> unit`: remove the account that has identifier `account_id` from the ledger,
   don't do anything if this account does not exist.
 
--- a/00xx-account-deletion.md
+++ b/00xx-account-deletion.md
@@ -301,22 +301,25 @@ Deleting an account $a₀$ results in two actions:
 
 Handling other tokens is the responsability of the smart contract(s) dealing with these tokens.
 
-Furthermore, account deletion should not be permitted except for some authorized
-participants, e.g., the contract that created the consumable account.
-
-We propose to restrict this capacity to the account to be deleted, so that only
-yourself can call for your own deletion.
-
-For that, we need to check this with respect to the `permissions` field of
-`Account_update.Update.t`.
-
-
 At the snark level, deletion should thus be equivalent to proving two things
 1. The location $l$ of account $a₀$ is now  equal to the empty account.
    A simple implementation is to call [`set_account`](https://github.com/MinaProtocol/mina/blob/4495af5caea5e1bb2f98f92592c065f93a586ade/src/lib/transaction_snark/transaction_snark.ml#L1531) to location $l$ with the empty account;
 2. The change in MINA of $a₁$ should be equal to
    - the initial creation fee;
    - augmented by any remaining MINA balance.
+
+
+**Permissions**. Furthermore, account deletion should not be permitted except
+for some authorized participants, e.g., the contract that created the consumable
+account. We will extend the [permissions
+type](https://github.com/MinaProtocol/mina/blob/4495af5caea5e1bb2f98f92592c065f93a586ade/src/lib/mina_base/permissions.ml#L343)
+with a dedicated `delete` field.  The handling of permissions will be similar as
+the one for [nonce
+update](https://github.com/MinaProtocol/mina/blob/4495af5caea5e1bb2f98f92592c065f93a586ade/src/lib/transaction_logic/zkapp_command_logic.ml#L1687)
+for the computation of `has_permission` and the updated `local_state`. This
+updated local state will also depend on extending [`Failure.t`](https://github.com/MinaProtocol/mina/blob/4495af5caea5e1bb2f98f92592c065f93a586ade/src/lib/mina_base/transaction_status.ml#L9) with an extra variant `Update_not_permitted_delete`.
+
+
 
 
 <!-- user command snark https://github.com/MinaProtocol/mina/blob/develop/src/lib/transaction_logic/mina_transaction_logic.ml#L1007
--- a/00xx-account-deletion.md
+++ b/00xx-account-deletion.md
@@ -431,7 +431,6 @@ column for all already registered rows.
 
 
 ### Other expected code changes
-#### Ledgers
 
 #### Removal
 
--- a/00xx-account-deletion.md
+++ b/00xx-account-deletion.md
@@ -159,13 +159,25 @@ There are some further details to be taken care of, such as:
 TBD: On-disk representation of the free list
 
 
-#### Ledger sync
+#### Syncing ledgers
 
-Now that the ledger is can have unfilled leaves
+The current implementation of [syncable
+ledger](https://github.com/MinaProtocol/mina/blob/4495af5caea5e1bb2f98f92592c065f93a586ade/src/lib/syncable_ledger/syncable_ledger.ml)
+assumes the allocated addresses are contiguous (see [this
+line](https://github.com/MinaProtocol/mina/blob/4495af5caea5e1bb2f98f92592c065f93a586ade/src/lib/syncable_ledger/syncable_ledger.ml#L301))
+when sending data.
 
-Encoding of the legder leaves:
-- Some(a) when there is an account a,
-- None if it´s empty
+A ledger supporting account deletion can have unfilled leaves, which breaks this assumption.
+Thus, we need to devise a suitable strategy to transfer this data.
+
+We propose to maintain the current syncing protocol with the following simple
+extension: send multiple chunks of contiguous blocks of allocated addresses instead of a single one that assumes no unallocated address.
+
+On the other side of the synchronization,
+[add_content](https://github.com/MinaProtocol/mina/blob/4495af5caea5e1bb2f98f92592c065f93a586ade/src/lib/syncable_ledger/syncable_ledger.ml#L408)
+needs to be handle this correctly and this in turn relies on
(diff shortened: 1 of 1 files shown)


00xx-account-deletion.md
```diff
@@ -0,0 +1,272 @@
...
+```
+
+Upon removal of `C`, the structure would evolve as follows, with `x` a
+placeholder value marking the freed location, which would actually represent to
+the empty account in practice.  The free list now states that the location
+determined by sequence of directions `[Right; Left]` is available for reuse for
+new insertions
+
+```
+    H(H(A,B),H(C,D))       free = [[Right; Left]]
```
mrmr1993, line 85: Indexes are already dense, shouldn't we use those?

mrmr1993 (changes requested):
I don't think this digs into the mechanisms in the way I was expecting. Can you add some detail about what needs to be implemented to achieve the described behaviour.

00xx-account-deletion.md
rbonichon: Do you mean using the `Bigstring` view of the same information, that is, `Merkle_address` ?

rbonichon: The actual implementation will use `Location.t`, which may or may not be  a compact representation since we are within the functor's body. The current instance of this functor does use a compact representation for indexes. I chose the `Right, Left` way of wording the location because I thought it was easier to follow, especially in a tree-like drawing.

mrmr1993 (changes requested):
Can you update this to be in line with [the template](https://github.com/o1-labs/rfcs/blob/main/0001-template.md)? More comments to follow

00xx-account-deletion.md
```diff
@@ -0,0 +1,352 @@
...
+
+```
+    H(H(A,B),H(C,D))               free = []
+      /       \
+   H(A,B)     H(C,D)
+   /  \       /  \
+  A    B     C    D
+```
+
+Upon removal of `C`, the structure would evolve as follows, with `x` a
```
mrmr1993: This is the only mention of this placeholder value. The details here seem important, e.g. we don't want a deleted account to continue to use the same amount of memory, otherwise we shouldn't refund the creation fee.

00xx-account-deletion.md
```diff
@@ -0,0 +1,352 @@
...
+*  account locations in this representation are fixed for the lifetime of the
+  account: proof of inclusion would not change but for the last element of the
+  Merkle path.
+
+
+#### On-disk ledger
+
+The `Database` module implements an on-disk ledger that backs up the in-memory data structure.
+
+Deletion support relies on the same techniques as in-memory ledgers, tailored to the on-disk db.
```
mrmr1993: What does this mean?

00xx-account-deletion.md
```diff
@@ -0,0 +1,352 @@
...
+The support for deletion at the ledger level now needs to be lifted within the
+transaction logic for account updates.
+The key idea here is to handle removal as a specific kind of account update for
+a given account.
+
+Concretely, the basic support is provided by adding a deletion flag on account
+updates, i.e., extend the `Account_update.Body.Stable.t` type with an additional
+record field `delete_account`.
+
+This turns the actual definition from
```
mrmr1993: Repeating the type twice here feels like redundant padding. Delete.

00xx-account-deletion.md
```diff
@@ -0,0 +1,352 @@
...
+        ; may_use_token : May_use_token.Stable.V1.t
+        ; authorization_kind : Authorization_kind.Stable.V1.t
+         (* if true, this account update will delete the account *)
+        ; delete_account: bool;
+        }
+    ```
+
+This extension provides is thus a version 2 of the current versioned type (`V2`).
+
+What about the changes in Mina_wire_Types?
```
mrmr1993: The transactions we send over the wire will change, surely

00xx-account-deletion.md
```diff
@@ -0,0 +1,352 @@
...
+
+Account deletion is also a payment of the initial account
+creation fee back to an address with extended with that triggers account
+deletion.
+
+Clearly, account deletion should not be permitted except for authorized keys, e.g., the
+contract that create the consumable account. For that, we need to check this
```
mrmr1993: Check what? Be concrete

00xx-account-deletion.md
```diff
@@ -0,0 +1,352 @@
...
+        ; delete_account: bool;
+        }
+    ```
+
+This extension provides is thus a version 2 of the current versioned type (`V2`).
+
+What about the changes in Mina_wire_Types?
+
+Account deletion is also a payment of the initial account
+creation fee back to an address with extended with that triggers account
```
mrmr1993: Garbled sentence

00xx-account-deletion.md
rbonichon: Indeed. Let me be more clear here.

00xx-account-deletion.md
```diff
@@ -0,0 +1,367 @@
...
+
+```
+    H(H(A,B),H(E,D))               free = []
+      /       \
+   H(A,B)     H(E,D)
+   /  \       /  \
+  A    B     E    D
+```
+
+##### Option 2: Maintain insertion on leftmost available location
```
mrmr1993: Alternative, dump it in the alternatives section.

00xx-account-deletion.md
```diff
@@ -0,0 +1,367 @@
...
+We opt to implement the tracking of freed locations.
+
+While this choice implies adding a structure in the masking ledger
+implementation to track freed locations, it offers 2 advantages
+
+* the cost of removal is slightly better since we do not need to update location
+   <-> account mappings;
+*  account locations in this representation are fixed for the lifetime of the
+  account: proof of inclusion would not change but for the last element of the
+  Merkle path.
```
mrmr1993: Alternatives section

00xx-account-deletion.md
```diff
@@ -0,0 +1,367 @@
...
+
+### Other expected code changes
+#### Ledgers
+
+#### Removal
+
+We propose to support removing elements in ledgers through  2 functions :
+
+- `val remove_location: t -> location -> unit`
+- `val remove_account: t -> account -> unit`
```
mrmr1993: What do these do?

00xx-account-deletion.md
```diff
@@ -0,0 +1,367 @@
...
+- updating Merkle paths on removal
+- updating the `all_accounts` function to avoid iterating over addresses that
+  have been removed
+
+
+### Transaction logic
+
+The support for deletion at the ledger level now needs to be lifted within the
+transaction logic for account updates.
+The key idea here is to handle removal as a specific kind of account update for
```
mrmr1993, line 385: There's no mention of the snark-related code here. That will be crucially important to get right.

00xx-account-deletion.md
rbonichon: Would you maybe prefer also to use symbol `o` in place of `x` since it actually is meant to mark the empty account

00xx-account-deletion.md
mrmr1993, line 385: How does the transaction logic differ from what's present today? E.g.
* How should the API to signal that we're deleting an account look?
* How should the native version of that work?
* How should the snark version of that work?
* What happens in the case of failure? Is it already handled?
* How is the global supply of Mina tracked faithfully?

mrmr1993 (changes requested):

00xx-account-deletion.md
rbonichon: Sure. I think this question (to myself actually) was more: could that be an issue?

00xx-account-deletion.md
rbonichon: Done

00xx-account-deletion.md
rbonichon: Done

00xx-account-deletion.md
rbonichon: Added specification

00xx-account-deletion.md
```diff
@@ -0,0 +1,384 @@
...
+a given account.
+
+Concretely, the basic support is provided by adding a deletion flag on account
+updates, i.e., extend the `Account_update.Body.Stable.t` type with an additional
+record field `delete_account`.
+
+This turns the current definition into
+    ```ocaml
```
mrmr1993: Indentation here breaks the markdown formatting.

00xx-account-deletion.md
```diff
@@ -0,0 +1,384 @@
...
+
+We will only discuss option 1 here and delay the discussion about option 2 in
+[the alternative section](#alternative).
+
+
+#### Example scenario
+
+To illustrate option 1, let us assume we have the following Merkle tree of depth
+2, with an empty free list. This Merkle tree has one location marked with `x` -
+this marks the empty account.
```
mrmr1993, line 81: I don't think this is the design we want in practice: if we just replace it with the empty account, we don't actually save any space. We would really like to delete the account entirely (at least, on disk).

00xx-account-deletion.md
```diff
@@ -0,0 +1,384 @@
...
+   H(A,B)     H(E,D)
+   /  \       /  \
+  A    B     E    D
+```
+
+##### Merkelizing the free list
+
+The view of the free list above is a simplification.
+
+The data contained in the free list needs to be Merkelized as well.  This means,
```
mrmr1993: > needs to be Merkelized as well

Please propose a usage if we're going to this effort. E.g. do we need to store it alongside the ledger hash now? Do we need to update the snark ledger update mechanism to account for it? How do we handle the empty merkle list in snark, where the account index is currently non-deterministic? How does this impact ledger sync?

00xx-account-deletion.md
```diff
@@ -0,0 +1,384 @@
...
+
+- Merkelization of ledger + free list and interface for interacting with this data structure
+
+
+## Resources - TB removed
+
+
+> Here are some things I am thinking about, maybe they help you inform the RFC:
+>
+> - easy API - we need to express account deletion via an easy API in the sdk
```
mrmr1993: The SDK should only be interfacing with the `Account_update.t` portion, so the API is 'handling and extra boolean' at that level, plus some wrapping TBD.

00xx-account-deletion.md
rbonichon: I see your point. We don't really need to store the empty account though, we just need to act as if it were stored in the leaf. I will rephrase that bit.

00xx-account-deletion.md
rbonichon: This part should be commented out. This was just for me to check that these points are addressed in the RFC

00xx-account-deletion.md
```diff
@@ -0,0 +1,471 @@
...
+the free list in that order, that the actual data in the free list would be:
+
+$$
+free = [ (loc_n, h_n = H(loc_n, h_{n-1})), ..., (loc_2, h_2 = H(loc_2, h_1)); (loc_1, h_1 = H(loc_1, nil))]
+$$
+
+In turn, the ledger state is the Merklized pair of the ledger itself (as a
+Merkle tree) and its free list.
+
+TBD: using an ordered free list to avoid Merklization
```
mrmr1993: Please add these details, and choose between the options, punting the other to the alternatives section.

00xx-account-deletion.md
```diff
@@ -0,0 +1,471 @@
...
+- the free list is identified by the key `free_list` (like [`last_account_location`](https://github.com/MinaProtocol/mina/blob/4495af5caea5e1bb2f98f92592c065f93a586ade/src/lib/merkle_ledger/database.ml#L235));
+- when parsing binary data associated to this key, we get two things:
+  1. the actual first free location within the Merkle tree;
+  2. the new value of the `free_list`.
+
+Basically, the free list is handled as a binary-encoded list with a lazy decoding function, a single element at a time.
+
+<!-- This needs to be rewritten, not sure this is correct -->
+We will implement a decoding function where one can specify how many locations
+we ideally want to retrieve (up to `n`) to handle functions `set_location_batch`.
```
mrmr1993: Don't we have to reserialise the whole thing anyway once we've removed these from the list? The data also grows linearly with the number of free'd accounts, can you justify why that's okay and/or propose an alternative?

mrmr1993 (commented):
Please add some details on how this affects the merkle masks.

00xx-account-deletion.md
```diff
@@ -0,0 +1,471 @@
...
+yourself can call for your own deletion.
+
+For that, we need to check this with respect to the `permissions` field of
+`Account_update.Update.t`.
+
+
+At the snark level, deletion should thus be equivalent to proving two things
+1. The location $l$ of account $a₀$ is now  equal to the empty account.
+   A simple implementation is to call [`set_account`](https://github.com/MinaProtocol/mina/blob/4495af5caea5e1bb2f98f92592c065f93a586ade/src/lib/transaction_snark/transaction_snark.ml#L1531) to location $l$ with the empty account;
+2. The change in MINA of $a₁$ should be equal to
```
mrmr1993: This is implicit in 1.

00xx-account-deletion.md
```diff
@@ -0,0 +1,471 @@
...
+
+Handling other tokens is the responsability of the smart contract(s) dealing with these tokens.
+
+Furthermore, account deletion should not be permitted except for some authorized
+participants, e.g., the contract that created the consumable account.
+
+We propose to restrict this capacity to the account to be deleted, so that only
+yourself can call for your own deletion.
+
+For that, we need to check this with respect to the `permissions` field of
```
mrmr1993: This sounds like you're adding a new permission to accounts. Please document.

00xx-account-deletion.md
```diff
@@ -0,0 +1,471 @@
...
+`Account_update.Update.t`.
```
mrmr1993: This record declares an update to the permissions, not the permissions themselves.

00xx-account-deletion.md
rbonichon: Updated

00xx-account-deletion.md
rbonichon: Fixed

00xx-account-deletion.md
rbonichon: The idea here would have been to deserialize/serialize only once instead of `n` times when do batch allocations.

00xx-account-deletion.md
rbonichon: Done

00xx-account-deletion.md
rbonichon: I'm not sure I get your point. How does setting the location of an account a to the empty account implicitly mean that another account b should get some amount of MINA?

00xx-account-deletion.md
rbonichon: Done

00xx-account-deletion.md
rbonichon: The last version addresses permissions.

00xx-account-deletion.md
rbonichon: Corrected

00xx-account-deletion.md
```diff
@@ -0,0 +1,576 @@
...
+**Imposing an ordering** from the structure of the data that we insert in the
+free list has better properties. Heaps, `OCaml` sets (which rely on a comparison
+function), or even ordered lists are implementation choices that would work in
+this context.
+
+We thus propose to use a **max-heap** (like
+[bheap](https://opam.ocaml.org/packages/bheap/) or
+[CCHeap](https://c-cube.github.io/ocaml-containers/last/containers/CCHeap/index.html))
+for this implementation, that is, the free list is an ordered set of locations,
+from which we always allocate the biggest one.
```
mrmr1993: Why are we choosing 'biggest' here? My intuition is that, especially if this feature is widely used, we might have accounts created and deleted at the right side of the tree; it would be natural to treat this as moving the fill frontier backwards. For example, if we had a series of deletes like
```
[a, b, c, d, e, f]
[a, b, c, X, e, f]
[a, b, c, X, e, X]
[a, b, c, X, X, X]
```
then it's natural to treat the states as
```
[a, b, c, d, e, f]
[a, b, c, X, e, f]
[a, b, c, X, e]
[a, b, c]
```
and recover the memory. However, in the current description, an insertion would give us
```
[a, b, c, X, X, g]
```
where in a sense we're always trying to 'waste' the maximum possible number of spaces by choosing the largest. If we instead choose the smallest, it really does behave like we've reduced the fill frontier, and always gives us the option to do that as a performance optimisation.

00xx-account-deletion.md
```diff
@@ -0,0 +1,576 @@
...
+
+- patch function [`allocate`](https://github.com/MinaProtocol/mina/blob/4495af5caea5e1bb2f98f92592c065f93a586ade/src/lib/merkle_ledger/database.ml#L267) to handle the free list as well
+- updating Merkle paths on removal
+- updating the `all_accounts` function to avoid iterating over addresses that
+  have been removed
+
+Other strategies are discussed [here](#free_list_db alt).
+
+
+#### Syncing ledgers
```
mrmr1993: As well as synchronising ledgers, we also need to consider node startup for genesis. Consider the `ledger` field in the runtime configuration: we need a way to represent that an account is empty in that format, and faithfully convert it, so that the genesis ledger after a hard-fork faithfully represents the state of the ledger once we have account deletion.

00xx-account-deletion.md
```diff
@@ -0,0 +1,576 @@
...
+module implements an on-disk ledger that backs up the in-memory data structure.
+
+Deletion support relies on the same techniques as in-memory ledgers, that is, a
+representation of a ledger as Merkle tree *and* an additional free list.
+
+The main culprit is the representation of the free list on-disk. Here we will simply store
+the data on the KV database as follows:
+- the free list is identified by the key `free_list` (like
+  [`last_account_location`](https://github.com/MinaProtocol/mina/blob/4495af5caea5e1bb2f98f92592c065f93a586ade/src/lib/merkle_ledger/database.ml#L235));
+- store the serialized encoding of the heap in the database.
```
mrmr1993: Please define this.

00xx-account-deletion.md
```diff
@@ -0,0 +1,576 @@
...
+
+**Allocation cost**. Allocating locations from the free list now means
+1. read `free_list`
+2. deserialize the heap value;
+3. get the needed free locations;
+4. write the accounts to the free locations;
+5. reserialize the new value of the heap.
+
+**Simple improvement on batch allocations**. Furthermore, we will implement a
+decoding function where one can specify how many locations we ideally want to
```
mrmr1993, line 320: Presumably this isn't useful in practise: usually, you'll be grabbing the free list so that you can update it to consume a location, so you'll need to reserialise the free list once the head elements (or tail elements as currently written here) have been removed.

It's also useful to mention performance: is there any risk that a large number of free'd accounts creates a free list that slows down accesses, because inherently we're writing a larger and larger list into the database as it grows?

00xx-account-deletion.md
```diff
@@ -0,0 +1,576 @@
...
+
+
+
+<!-- user command snark https://github.com/MinaProtocol/mina/blob/develop/src/lib/transaction_logic/mina_transaction_logic.ml#L1007
+ !-- zkapp command snark https://github.com/MinaProtocol/mina/blob/develop/src/lib/transaction_logic/zkapp_command_logic.ml#L1232 -->
+
+### Updating the archive node
+
+The addition of a field in `Account_update` will need to reflected in table
+`zkapp_account_update_body` with the addition of a `delete_account` column.
```
mrmr1993, line 515: Add details for the permission.

00xx-account-deletion.md
```diff
@@ -0,0 +1,576 @@
...
+#### Ledgers
+
+#### Removal
+
+We propose to support removing elements in ledgers through  2 functions, that we specify below:
+
+- `val remove_location: t -> location -> unit`: remove the account found at the
+  given `location` from the ledger, don't do anything if `location` is occupied
+  by the empty account
+- `val remove_account: t -> account -> unit`: remove `account` from the ledger,
```
mrmr1993: What is `account` here? `account_id` perhaps? Or do you pass a full account that might be in the ledger?

00xx-account-deletion.md
rbonichon: I explain this in the example right below in the document. There are 2 elements for allocations in the structure:
1. The free list (indices between [0, fill frontier index[), let's call it l
2. The fill frontier index, let's call it i

The strategy would be to always use the free list first (from the biggest to the lowest index), then the fill frontier index if the the free list is empty. 

Replaying your example, this would give 
```
[a, b, c, d, e, f]       l = [] i = 5
[a, b, c, X, e, f]      l = [3]  i = 5
[a, b, c, X, e, X]     l = [3] i = 4
[a, b, c, X, X, X]     l = []  i = 2 ( this is where it's important to have a max heap, to make this reduction easy, since you just have to peek and pop)
```
Thus the next insertion would give 
`[ a, b, c, g]`as you described since the free list has been emptied.

00xx-account-deletion.md
rbonichon: We do use batch allocations in the code. To be sure, I don't know whether the code is executed in practice.

00xx-account-deletion.md
rbonichon: It's an abstract `account` type as defined [here](https://github.com/MinaProtocol/mina/blob/4495af5caea5e1bb2f98f92592c065f93a586ade/src/lib/merkle_ledger/intf.ml#L278). 

Come to think of it an `account_id`might be better.

rbonichon: Updated

00xx-account-deletion.md
rbonichon: The scenario you describe was supposed to be addressed by the paragraph starting with *Deallocation at fill frontier*

00xx-account-deletion.md
rbonichon: Done

00xx-account-deletion.md
rbonichon: Done

00xx-account-deletion.md
mrmr1993: We do, but the point is that those batch allocations immediately result in a write of the new data. That is to say: deserialising just the head(s) is usually not what you want, given that you also want the tail so that you can reserialise it.

00xx-account-deletion.md
```diff
@@ -0,0 +1,712 @@
...
+
+This simple change trickles down into updating the transitive closure of all
+type definitions depending on `Account_update`, including
+`Mina_wire_types.Mina_transaction`.
+
+#### Effect of deleting an account
+
+Deleting an account $a₀$ results in two actions:
+1. Actual removal of account $a₀$ of the storage layer;
+2. Return of balance of MINA tokens from $a₉$ and initial creation fee to a specified $a₁$ account.
```
mrmr1993: This isn't how we want it in practice. Instead, we want to return the balance to the 'floating balance' for the zkApp txn, and the rest of the txn logic handles the other account. We *should not* specify the `a_1` account (and indeed that might be several accounts in actual usage).

00xx-account-deletion.md
```diff
@@ -0,0 +1,712 @@
...
+2. The change in MINA of $a₁$ should be equal to
+   - the initial creation fee;
+   - augmented by any remaining MINA balance.
+
+
+**Permissions**. Furthermore, account deletion should not be permitted except
+for some authorized participants, e.g., the contract that created the consumable
+account. We will extend the [permissions
+type](https://github.com/MinaProtocol/mina/blob/4495af5caea5e1bb2f98f92592c065f93a586ade/src/lib/mina_base/permissions.ml#L343)
+with a dedicated `delete` field.
```
mrmr1993: Default value?

00xx-account-deletion.md
```diff
@@ -0,0 +1,712 @@
...
+
+### Storage
+
+<!-- sparse ledger https://github.com/MinaProtocol/mina/blob/develop/src/lib/mina_base/sparse_ledger_base.ml#L85
+ !-- mask ledger https://github.com/MinaProtocol/mina/blob/develop/src/lib/merkle_mask/masking_merkle_tree.ml#L973
+ !-- db ledger https://github.com/MinaProtocol/mina/blob/develop/src/lib/merkle_ledger/database.ml#L559 -->
+
+Within the protocol-related codebase, the lowest-level support lies in having a
+working and sound `remove` functions for the different ledger implementations
+(`Database`, `Any_ledger`, `Null_ledger`, `Syncable_ledger`) and in masking
```
mrmr1993, line 35: I think we need more info here about `masking merkle trees`. In particular, it's important to specify the `commit` method to ensure that we're atomically updating all of the relevant information when we e.g. push to disk or apply a zkApp transaction.

00xx-account-deletion.md
mrmr1993: There are 2 things in conflict here: coalescing the free space to the right while also inserting to the left to keep it dense. You've addressed one, I was addressing the other. Please talk to both in the RFC. It seems as though either option probably require us to deserialise the whole list (unless we have head and bottom peeking).

00xx-account-deletion.md
rbonichon: `Signature`. Done

00xx-account-deletion.md
rbonichon: Addressed

00xx-account-deletion.md
rbonichon: I tried to explain why it might be useful. We do have to serialize some data back to be written (the tail). For batch allocations, I think it's useful to be able to get `n` locations at once to avoid `n-1` read/deserialize/serialize/write` sequences.

00xx-account-deletion.md
rbonichon: Creating then deleting an account at the fill frontier is already correctly handled [here](https://github.com/MinaProtocol/mina/blob/4495af5caea5e1bb2f98f92592c065f93a586ade/src/lib/merkle_mask/masking_merkle_tree.ml#L497) by just moving the `current_location` (what I call the fill frontier). My understanding is this is what you mean by "coalescing the free space to the right".  

I think this step in the example 
```
[a, b, c, X, e, f]      l = [3]  i = 5
[a, b, c, X, e, X]     l = [3] i = 4
```
coalesces the free space to the right (i moves from 5 to 4, the free list, i.e, the left,  is not updated)
This other step 
```
[a, b, c, X, e, X]     l = [3] i = 4
[a, b, c, X, X, X]     l = []  i = 2
```
shows how the free list and the fill frontier interact with one another to keep the free space to the right ¨coalesced¨. 

I'm not sure what I am missing. Could you show me another example of where what I am proposing doesn't offer a good solution?

mrmr1993:
```
[a, b, c, X, X, f, X]     l = [4, 3] i = 4
[a, b, c, X, e, f, X]     l = [3]  i = 2
```
This is bad.

00xx-account-deletion.md
```diff
@@ -0,0 +1,732 @@
...
+date. Now, when getting an account (see
+[get](https://github.com/MinaProtocol/mina/blob/4495af5caea5e1bb2f98f92592c065f93a586ade/src/lib/merkle_mask/masking_merkle_tree.ml#L248)),
+the child only has to check whether the location has been removed before sending
+the request to its parent.
+
+**Committing** a masking tree to its parent is impacted by removal as well. [The
+commit
+function](https://github.com/MinaProtocol/mina/blob/4495af5caea5e1bb2f98f92592c065f93a586ade/src/lib/merkle_mask/masking_merkle_tree.ml#L617)
+will need to push its removed locations to its parent atomically, that is, after
+the `is_committing` flage has been set to true
```
mrmr1993: `s/flage/flag`

00xx-account-deletion.md
mrmr1993: I agree, but at a certain point it's simpler to deserialise the whole list, take as many locations as are needed, push any new locations onto the list, and then reserialise the final list.

00xx-account-deletion.md
rbonichon: No in your example `i` would still be 4. Allocating from the free list does not change the fill frontier.

rbonichon: In reverse, deallocation from the fill frontier might affect the free list, like in the example above

rbonichon: I will add a section on allocation strategy to make that clearer. It does not seem to be.

rbonichon: The contents of this discussion is reflected in the document

00xx-account-deletion.md
rbonichon: I agree too. That brings the following question: Is there a scenario where we would in one call allocate mutliple and free multiple locations at the same time?

00xx-account-deletion.md
rbonichon: Fixed

00xx-account-deletion.md
nholland94: What @mrmr1993 says here is correct. Luckily, it's very easy (in fact, easier than what is described in the RFC as of now): you just need to track the initial creation fee in the fee excess, no other changes will be required.

rbonichon: Updated the document to reflect this discussion

PR #52, opened 2024-07-25
XT Example App: Name Registry using off chain state

45930:
[Link to rendered RFC](https://github.com/o1-labs/rfcs/blob/76045f062f87c4ab98c95ad7bed3bddb0cd565ed/00xx-offchain-state-showcase.md)

I don't know if an RFC is the right approach to get this design approved, but we will need some process for external review of our designs.  The goal is to align our implementations of the SDK with the company's strategic goals and show off how things are "meant to be" used.  I am leaving this as a draft for now, pending clarity on the process.


45930:
I chatted with DFST about this design, and his design for the minascan solution.

He has also noted the dependency on archive nodes as being an issue:

> I do not plan to use now new off-chain state in o1js as it is designed now as instability of archive node while processing actions should be resolved first.

